PR #1, opened 2015-02-06
Silently ignore unknown permissions instead of throwing an IllegalArgumentException

cyrille-leclerc:
Better exception message if a permission does not exist.

The current message does not indicate which permission is unknown:

```
Caused by: java.lang.IllegalArgumentException
    at hudson.security.GlobalMatrixAuthorizationStrategy.add(GlobalMatrixAuthorizationStrategy.java:94)
    at hudson.security.GlobalMatrixAuthorizationStrategy$DescriptorImpl.newInstance(GlobalMatrixAuthorizationStrategy.java:276)
    at hudson.security.GlobalMatrixAuthorizationStrategy$DescriptorImpl.newInstance(GlobalMatrixAuthorizationStrategy.java:248)
```

New message

```
Caused by: java.lang.IllegalArgumentException: Unknown permission:"com.cloudbees.plugins.updatecenter.UpdateCenter.Configure" set to "true" for sid "configure_jenkins"
	at hudson.security.GlobalMatrixAuthorizationStrategy$DescriptorImpl.newInstance(GlobalMatrixAuthorizationStrategy.java:279)
	at hudson.security.GlobalMatrixAuthorizationStrategy$DescriptorImpl.newInstance(GlobalMatrixAuthorizationStrategy.java:250)
```

### DIFF
--- a/src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
+++ b/src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
@@ -276,9 +276,10 @@ public class GlobalMatrixAuthorizationStrategy extends AuthorizationStrategy {
                     if ((Boolean) e.getValue()) {
                         Permission p = Permission.fromId(e.getKey());
                         if (p == null) {
-                            throw new IllegalArgumentException("Unknown permission:\"" + e.getKey() + "\" set to \"" + e.getValue() +"\" for sid \"" + sid + "\"");
+                            LOGGER.log(Level.FINE, "Silently skip unknown permission \"{0}\" for sid:\"{1}\"", new Object[]{e.getKey(), sid});
+                        } else {
+                            gmas.add(p, sid);
                         }
-                        gmas.add(p,sid);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
```diff
@@ -273,6 +275,9 @@ public AuthorizationStrategy newInstance(StaplerRequest req, JSONObject formData
                     }
                     if ((Boolean) e.getValue()) {
                         Permission p = Permission.fromId(e.getKey());
+                        if (p == null) {
+                            throw new IllegalArgumentException("Unknown permission:\"" + e.getKey() + "\" set to \"" + e.getValue() +"\" for sid \"" + sid + "\"");
+                        }
```
jglick: This is wrong. It should simply skip over the nonexistent permission; it is a normal condition, for example when you disable a plugin you had previously been using. At most the condition should be logged at `FINE`.

jglick:
:+1:

KostyaSha:
No tests!

jglick:
Yeah a test would have been nice here.

KostyaSha:
I can copy this test https://github.com/jenkinsci/build-pipeline-plugin/pull/72#discussion_r29333536 :)

PR #2, opened 2015-07-16
Enable FindBugs for the plugin

oleg-nenashev:
@reviewbybees

### DIFF
--- a/src/main/java/hudson/security/AuthorizationMatrixProperty.java
+++ b/src/main/java/hudson/security/AuthorizationMatrixProperty.java
@@ -58,6 +58,8 @@ import com.thoughtworks.xstream.converters.MarshallingContext;
 import com.thoughtworks.xstream.converters.UnmarshallingContext;
 import com.thoughtworks.xstream.io.HierarchicalStreamReader;
 import com.thoughtworks.xstream.io.HierarchicalStreamWriter;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import javax.annotation.CheckForNull;
 
 import javax.servlet.ServletException;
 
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
Te code formatting in the plugin is horrible BTW.

src/main/java/hudson/security/AuthorizationMatrixProperty.java
```diff
@@ -300,9 +306,9 @@ public Object unmarshal(HierarchicalStreamReader reader,
 				reader.moveDown();
 				reader.getValue(); // we used to use this but not any more.
 				reader.moveUp();
-				prop = reader.peekNextChild();
+				prop = reader.peekNextChild(); // We check the next field
 			}
-			else if ("blocksInheritance".equals(prop)) {
```
stephenc: Seems like a non-findbugs related change... not saying it is incorrect... rather that it seems like one

oleg-nenashev: It's a FindBugs one. Otherwise FindBugs complains that the assigned value from the previous if clause is never used. Probably it should be registered as a separate functional bug, but I was not able to find related issues

stephenc: Sounds like something that should be merged but in a separate pull request then

oleg-nenashev: Agreed

oleg-nenashev:
@stephenc 
Created https://issues.jenkins-ci.org/browse/JENKINS-29527 and removed the change from this PR

stephenc:
:bee:

oleg-nenashev:
@reviewbybees done

PR #3, opened 2015-07-24
[FIXED JENKINS-29527] - AuthorizationMatrixProperty converter: Inheritance blocking may be passed incorrectly

oleg-nenashev:
It's a follow-up to #2 

If "useProjectSecurity" appears in the marshalled stream before "blocksInheritance", the second option may be ignored => the project may start taking inherited permissions. However, seems the code just iterates through fields => we should not use the else clause. The assigned "prop" variable is never used, so it's another clue of the bug

@reviewbybees @stephenc


varmenise:
:bee:

oleg-nenashev:
@reviewbybees done

jglick:
A test would have been in order I think.

PR #4, opened 2015-07-27
Create README.md

gliptak:


oleg-nenashev:
I'd recommend to remove this link. It performs builds of pull requests, so the public status may become red if somebody submits a wrong PR

gliptak:
Is there a different link pointing to master branch in main repository? Thanks

oleg-nenashev:
@gliptak 
We presume that everybody uses pull requests to commit changes, hence the master branch is always green. If somebody violates the process and the build fails, committers and plugin maintainers get notifications from pull request builders. So we don't expect a long "failed status" as well.

In general it would be useful to have a direct link to the master branch. I'd recommend to create a JIRA issue in the INFRA project and describe the cases you see there. After that somebody from ci.jenkins-ci.org or jenkins.ci.cloudbees.com admins may take the task and implement it.

jglick:
Yeah I would suggest skipping the status icon too, it will usually be misleading.

jglick:
(There are plans afoot to replace the current plugin builds with a multibranch system on `ci.jenkins-ci.org`, which would not suffer from this issue.)

gliptak:
@jglick Could you point me to the updated URL format? Thanks

PR #5, opened 2015-07-27
Note case sentivity in help

gliptak:


oleg-nenashev:
:+1:

gliptak:
Anything else needed before this can be committed? Thanks

oleg-nenashev:
It would make sense to replace "authentication backend" to "security realm" just to reference common Jenkins terms. You could also add a hyperlink to https://wiki.jenkins-ci.org/display/JENKINS/Securing+Jenkins in order to clarify it

daniel-beck:
I don't think this is correct. Is there a security realm this does not apply to?

gliptak:
@oleg-nenashev Updated text per your comments
@daniel-beck I'm not sure if it does or doesn't apply to all, but the mention of "case sensitivity" likely will help some others out

jglick:
Is this still applicable given #9?

gliptak:
I would like to deploy an updated plugin and check ...

PR #7, opened 2015-09-20
Improve the formatting of the validation responses

daniel-beck:
* All regular responses now have a tooltip on the icon indicating what was (not) found
* Different icon when no user/group could be found
* When no user/group found, text is grey and struckthrough

> screenshot


jglick:
:+1:

src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
```diff
@@ -363,16 +363,24 @@ public FormValidation doCheckName_(@Nonnull String value, @Nonnull AccessControl
                 }
 
                 // couldn't find it. it doesn't exist
-                return FormValidation.respond(Kind.ERROR, makeImg("error.png") +ev);
+                return FormValidation.respond(Kind.ERROR, makeImg("user-disabled.png", "User or group not found", true) + formatNonexistentUser(ev));
```
jglick: :ant: Should make these localizable at some point.

PR #8, opened 2015-12-02
[JENKINS-31860] Restore return type to match API

stephenc:
https://issues.jenkins-ci.org/browse/JENKINS-31860

@reviewbybees 

I am not entirely happy with this approach as when we have one of the parent / child ACLs being non `SidACL` then the semantics of `Boolean hasPermission(Sid p, Permission permission)` may be subverted... IOW we have no way to differentiate between the Child ACL saying "`Boolean.FALSE`->no and don't check the parent" and the Child ACL saying "`null`->only if parent is ok"

### DIFF
--- a/src/main/java/hudson/security/ProjectMatrixAuthorizationStrategy.java
+++ b/src/main/java/hudson/security/ProjectMatrixAuthorizationStrategy.java
@@ -35,6 +35,8 @@ import com.thoughtworks.xstream.io.HierarchicalStreamReader;
 import com.thoughtworks.xstream.converters.UnmarshallingContext;
 import com.thoughtworks.xstream.mapper.Mapper;
 import com.thoughtworks.xstream.core.JVM;
+import org.acegisecurity.Authentication;
+import org.acegisecurity.acls.sid.Sid;
 import org.jenkinsci.plugins.matrixauth.Messages;
 
 import java.util.HashSet;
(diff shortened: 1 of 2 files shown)


jglick:
I am not sure what the concrete impact would be, but it seems like an API design flaw in core. Perhaps the three-way return type needs to be exposed in `ACL` itself (and `public` not `protected`)?

jglick:
Really deserves a test in `branch-api` or downstream (existing related test code is in `workflow-multibranch` for lack of adequate mocks).

jglick:
But :bee: so far as it goes—from code inspection, ought to solve the problem well enough. I think the impact of not handling null is at least safe, since `SidACL` defaults to rejecting access.

src/main/java/hudson/security/ProjectMatrixAuthorizationStrategy.java
```diff
@@ -56,16 +58,29 @@ public ACL getACL(Job<?,?> project) {
...
+                return projectAcl;
             }
-
-            return projectAcl;
         } else {
             return getACL(project.getParent());
         }
     }
 
-    public SidACL getACL(ItemGroup g) {
```
jglick: Unfortunately this is incompatible for `cloudbees-folder`:

```
java.lang.NoSuchMethodError: hudson.security.ProjectMatrixAuthorizationStrategy.getACL(Lhudson/model/ItemGroup;)Lhudson/security/SidACL;
	at com.cloudbees.hudson.plugins.folder.AbstractFolder.getACL(AbstractFolder.java:530)
```

We need to invert the dependency here.

stephenc: Or maybe add `@WithBridgeMethods`?

src/main/java/hudson/security/ProjectMatrixAuthorizationStrategy.java
```diff
@@ -56,16 +58,29 @@ public ACL getACL(Job<?,?> project) {
...
+    private static ACL inheritingACL(final ACL parent, final ACL child) {
+        if (parent instanceof SidACL && child instanceof SidACL) {
+            return ((SidACL) parent).newInheritingACL((SidACL) child);
```
jglick: cf. #10

src/main/java/hudson/security/ProjectMatrixAuthorizationStrategy.java
jglick: Irrelevant if https://github.com/jenkinsci/cloudbees-folder-plugin/pull/33 is accepted.

PR #10, opened 2016-02-08
Add cloudbees-folder dependency

jglick:
Counterpart to https://github.com/jenkinsci/cloudbees-folder-plugin/pull/33, and would have to be released simultaneously to minimize the chance of loss of functionality. (I do not know of a way to _force_ users who upgrade `cloudbees-folder` to also upgrade `matrix-auth`, alas.)

@reviewbybees esp. @stephenc

PR #11, opened 2016-03-16
Make extension for Folders plugin optional

daniel-beck:
Because https://github.com/jenkinsci/matrix-auth-plugin/blob/master/pom.xml#L35

Untested.

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
@@ -132,7 +132,7 @@ public class AuthorizationMatrixProperty extends AbstractFolderProperty<Abstract
         sids.add(sid);
     }
 
-    @Extension
+    @Extension(optional = true)
     public static class DescriptorImpl extends AbstractFolderPropertyDescriptor {
         @Override
         public AbstractFolderProperty<?> newInstance(StaplerRequest req, JSONObject formData) throws FormException {


daniel-beck:
@jglick WDYT? Would make 2.0 setup much nicer…

daniel-beck:
@reviewbybees Anyone?

jglick:
:bee: provided that you tested running with the dependency disabled and confirmed that you got only the single-line warning, not the stack trace. (Predicting the conditions under which you will get a `LinkageError` is rather tricky. Hence [JENKINS-19508](https://issues.jenkins-ci.org/browse/JENKINS-19508).)

daniel-beck:
@jglick I'm not getting any warning anymore. Steps to reproduce: 

* Install 2.0-beta-1, select to install `config-file-provider` (to get `icon-shim`) and finish setup
* Upload the PR build of this to your instance (I used my local build, but same content)
* Restart

No warning from `matrix-auth` before or after restart. It confirmed I can choose Matrix security, i.e. it's loading successfully. Folders plugin isn't installed. Not quite the same code path as installation via an update site, but I think warnings should appear either way.

jglick:
Instead of

```
… hudson.ExtensionFinder$Sezpoz _find
WARNING: Failed to load com.cloudbees.hudson.plugins.folder.properties.AuthorizationMatrixProperty$DescriptorImpl
java.lang.InstantiationException: java.lang.NoClassDefFoundError: com/cloudbees/hudson/plugins/folder/AbstractFolderPropertyDescriptor
	at net.java.sezpoz.IndexItem.element(IndexItem.java:146)
	at …
```

you should see a one-line warning (no stack trace)

    Failed to instantiate optional component com.cloudbees.hudson.plugins.folder.properties.AuthorizationMatrixProperty$DescriptorImpl; skipping

from `FaultTolerantScope`.

daniel-beck:
@jglick Have you tried this with the PR build? Because I have, as described, and it does *not* show a warning.

daniel-beck:
Would appreciate a merge and release here so Jenkins 2 doesn't come up with ugly stack traces in the default configuration.

jglick:
Intended to fix mistake in #10, and confirmed that it suppresses warnings in core 2.7-SNAPSHOT. I am not sure at what point `FaultTolerantScope` stopped printing its one-line warning for such cases, but at any rate that is probably a good thing.

PR #12, opened 2016-05-13
Fixes PCT against 2.x baselines.

armfergom:
Bumping parent pom version fixes PCT against 2.x baselines.

@reviewbybees


andresrc:
:bee:

PR #13, opened 2016-11-01
Fixes invisible text in .caption-row on #hudson-security-ProjectMatrixAuthorizationStrategy

jnbnyc:
Fixes invisible text in .caption-row on hudson-security-ProjectMatrixAuthorizationStrategy from font-color #f5f5f5!important on white background

Before:
screen shot 2016-11-01 at 1 58 18 pm

After:
screen shot 2016-11-01 at 2 04 19 pm


jnbnyc:
Disregard this, I just saw that the simple-theme-plugin is installed and the custom theme applied is causing the error.

PR #15, opened 2017-01-15
[JENKINS-39873] Let subprojects/subfolders choose to ignore parent folder permissions

wouterhund:
The current behavior remains unchanged, but this change adds a toggle to the project and folder configurations which will have that project/folder not inherit the permissions of its parent folder. It will still inherit from the global acl.

When you have a folder structure such as:

* a/b1/foo1
* a/b2/foo2

You can then grant everyone read permission on folder a, grant a certain group read permission on a/b1 which is not set to inherit from the parent folder (but still inherit from the global authorization matrix), then do the same for another group on a/b2. This only lets the groups see their own folders and contents.

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
@@ -79,6 +79,8 @@ public class AuthorizationMatrixProperty extends AbstractFolderProperty<Abstract
     private final Map<Permission, Set<String>> grantedPermissions = new HashMap<Permission, Set<String>>();
 
     private Set<String> sids = new HashSet<String>();
+    
+    private boolean blocksParentInheritance = false;
 
     protected AuthorizationMatrixProperty() {
     }
@@ -141,6 +143,10 @@ public class AuthorizationMatrixProperty extends AbstractFolderProperty<Abstract
                 return null;
 
             AuthorizationMatrixProperty amp = new AuthorizationMatrixProperty();
+            
+            // Disable parent inheritance, if so configured
+            amp.setBlocksParentInheritance(!formData.getJSONObject("blocksParentInheritance").isNullObject());
+            
             for (Map.Entry<String, Object> r : (Set<Map.Entry<String, Object>>) formData.getJSONObject("data").entrySet()) {
                 String sid = r.getKey();
                 if (r.getValue() instanceof JSONObject) {
@@ -204,6 +210,29 @@ public class AuthorizationMatrixProperty extends AbstractFolderProperty<Abstract
     public SidACL getACL() {
         return acl;
     }
+    
+    /**
+     * Sets the flag to block parent inheritance
+     *
+     * @param blocksParentInheritance
+     *            true if the authorization matrix should block inheritance from
+     *            parent
+     */
+    protected void setBlocksParentInheritance(boolean blocksParentInheritance) {
+        this.blocksParentInheritance = blocksParentInheritance;
+    }
+
+    /**
+     * Returns true if the authorization matrix is configured to block
+     * inheritance from the parent.
+     *
+     * @return true if the authorization matrix is configured to block
+     *         inheritance from parent
+     * @since TODO
+     */
+    public boolean isBlocksParentInheritance() {
+        return this.blocksParentInheritance;
+    }
 
     /**
      * Checks if the given SID has the given permission.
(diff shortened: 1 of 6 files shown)


wouterhund:
I'm not a cloudbees employee, but since this touches cloudbees code I'm hoping they will be kind enough to review this change. @reviewbybees

jglick:
> since this touches cloudbees code

This is not CloudBees code. In fact the CloudBees product does not use this plugin at all.

wouterhund:
> This is not CloudBees code. In fact the CloudBees product does not use this plugin at all.

I assumed due to a package being named `com.cloudbees`, and some other PRs made by cloudbees people. My bad!

daniel-beck:
Isn't it time the permissions be properly made tri-state: Grant / Deny / Inherit? Otherwise the next PR will be to ignore permissions of grandparent items.

daniel-beck:
@jenkinsci/code-reviewers via https://groups.google.com/d/msg/jenkinsci-dev/75x-_ZSaUFM/IahiVxAnBAAJ

wouterhund:
@daniel-beck I agree that would be superior. I already have a change in the pipeline (not ready to merge, uses reflection, can be cleaned up) which makes individual permissions non-inheritable.

I am interested in such a change as well, but since this is my first Jenkins contribution I wanted to keep it small and simple :)

hussem1:
When this will be merged?
Thank you...

adrianlzt:
Any news?

src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
```diff
@@ -204,6 +210,25 @@ protected Boolean hasPermission(Sid sid, Permission p) {
...
+     */
+    protected void setBlocksParentInheritance(boolean blocksParentInheritance) {
+        this.blocksParentInheritance = blocksParentInheritance;
+    }
+
+    /**
+     * Returns true if the authorization matrix is configured to block
+     * inheritance from the parent.
+     *
+     * @return
```
oleg-nenashev: Nice to have: description for return and `@since TODO`

src/main/java/hudson/security/AuthorizationMatrixProperty.java
```diff
@@ -216,7 +221,26 @@ public SidACL getACL() {
 	}
 
 	/**
-	 * Sets the flag to block inheritance
+	 * Sets the flag to block parent inheritance
+	 *
+	 * @param blocksParentInheritance
```
oleg-nenashev: Missing Javadoc

src/main/java/hudson/security/AuthorizationMatrixProperty.java
```diff
@@ -216,7 +221,26 @@ public SidACL getACL() {
...
+	 */
+	protected void setBlocksParentInheritance(boolean blocksParentInheritance) {
+		this.blocksParentInheritance = blocksParentInheritance;
+	}
+
+	/**
+	 * Returns true if the authorization matrix is configured to block
+	 * inheritance from the parent.
+	 *
+	 * @return
```
oleg-nenashev: Missing Javadoc

src/main/java/hudson/security/AuthorizationMatrixProperty.java
```diff
@@ -312,6 +342,13 @@ public Object unmarshal(HierarchicalStreamReader reader,
 			    reader.moveDown();
 			    as.setBlocksInheritance("true".equals(reader.getValue()));
 			    reader.moveUp();
+                            prop = reader.peekNextChild(); // We check the next field
+			}
+
+			if("blocksParentInheritance".equals(prop)) {
```
oleg-nenashev: Such approach presumes strict order of fields. It is generally OK, but having the logic within `while` iterator would be preferable. Not an issue in this PR

wouterhund:
@oleg-nenashev I've updated the PR with javadoc added.

src/main/java/hudson/security/AuthorizationMatrixProperty.java
wouterhund: Updated but github still shows

jglick:
Needs merge conflict resolution.

daniel-beck:
I have some work in progress that replaces this approach with a more flexible one (which also deprecates the existing `blocksInheritance`).

daniel-beck:
I believe https://github.com/jenkinsci/matrix-auth-plugin/pull/15 supersedes this. PTAL.

daniel-beck:
@wouterhund Please see previous comment.

daniel-beck:
Closing this PR in the expectation that https://github.com/jenkinsci/matrix-auth-plugin/pull/22 will make it into the next release (which will happen unless someone objects).

PR #16, opened 2017-02-21
[JENKINS-29815] Add disabling of permission inheritance for Folders

mathias-nyman:
This adds the same tick-box to disable inheritance of global permissions
to Folders as already exists for Projects. Consistency!

Support for the Folder plugin seems to originally have been done by
copy-pasting in an only slightly modified version of an ancient copy of
the AuthorizationMatrixProperty.java file. This commit is as such very
similar to the patch Kohsuke once applied to the Project's AMP:
6411fffb41e4e7b217d0e841462ef23a8e62c45a

Now, these two files are close to identical, and should be refactored
into one class in my opinion:

    src/main/java/hudson/security/AuthorizationMatrixProperty.java
    src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
@@ -80,6 +80,8 @@ public class AuthorizationMatrixProperty extends AbstractFolderProperty<Abstract
 
     private Set<String> sids = new HashSet<String>();
 
+    private boolean blocksInheritance = false;
+
     protected AuthorizationMatrixProperty() {
     }
 
@@ -141,6 +143,10 @@ public class AuthorizationMatrixProperty extends AbstractFolderProperty<Abstract
                 return null;
 
             AuthorizationMatrixProperty amp = new AuthorizationMatrixProperty();
+
+            // Disable inheritance, if so configured
+            amp.setBlocksInheritance(!formData.getJSONObject("blocksInheritance").isNullObject());
+
             for (Map.Entry<String, Object> r : (Set<Map.Entry<String, Object>>) formData.getJSONObject("data").entrySet()) {
                 String sid = r.getKey();
                 if (r.getValue() instanceof JSONObject) {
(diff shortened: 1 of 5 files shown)


mathias-nyman:
@wouterhund, I saw that you have a similar endeavour going on in https://github.com/jenkinsci/matrix-auth-plugin/pull/15. But to me it looked like you would be able to rebase your change on top of this one, which merely adds consistency to how Project and Folder permissions can be configured right now. Thoughts?

src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
```diff
@@ -265,6 +296,20 @@ public Object unmarshal(HierarchicalStreamReader reader,
                 final UnmarshallingContext context) {
             AuthorizationMatrixProperty as = new AuthorizationMatrixProperty();
 
+			String prop = reader.peekNextChild();
```
wouterhund: Indentation seems off

wouterhund:
Looks like it's similar indeed. Unfortunately my pull request doesn't seem to have much (or any) traction towards getting merged so far.

What do you propose? For one to rebase on the other and turn it into a single PR, or just see which one gets merged first and then have the other rebase on top of that.

src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
mathias-nyman: Thanks, good find! It's because there are tabs there, and not spaces. The other AuthorizationMatrixProperty file I copy-pasted this from has a mishmash of tabs and spaces :(.

I'll push a new patchset which uses only spaces.

mathias-nyman:
@wouterhund, yea I noticed the lack of enthusiasm for your PR https://github.com/jenkinsci/matrix-auth-plugin/pull/15. That's why I made this standalone PR which merely fixes an inconsistency bug in my opinion.

I support the idea of refactoring the way permission inheritance is configured, and that should only be easier after this PR is merged, as there's less functional change then while refactoring.

So, keeping this and PR https://github.com/jenkinsci/matrix-auth-plugin/pull/15 makes sense in my opinion. Either way it's quite a trivial rebase, so let's just see which one wins the race :)

wouterhund:
Sounds good.

> I support the idea of refactoring the way permission inheritance is configured, and that should only be easier after this PR is merged, as there's less functional change then while refactoring.

I certainly do as well, but don't have the time to dig into this currently. For now my needs are satisfied with my own fork, but hopefully either of these PRs gets through and then we'll see.

PR #9, opened 2015-12-02
[JENKINS-23805] Add support for the security realm's provided case sensitivity

stephenc:
[JENKINS-23805](https://issues.jenkins-ci.org/browse/JENKINS-23805)

@reviewbybees

(this was sitting in my workspace for who knows how long)

PR #18, opened 2017-09-05
[JENKINS-46654] Add node property for matrix auth

daniel-beck:
Plus Java 8 modernization (default methods save so many lines of code here, and could have prevented diverged behavior in https://github.com/jenkinsci/matrix-auth-plugin/pull/9#issuecomment-327262159).

> screen shot

Label is currently inconsistent with other configurations as this uses the display name.


daniel-beck:
http://stats.jenkins.io/pluginversions/matrix-auth.html:

For plugin versions 1.5 and newer (i.e. those released in the last 5 months), 86%/90%/94% of instances are on 2.46.1 or newer. So this does not impact many users who actually upgrade their plugins.

pom.xml
```diff
@@ -12,8 +12,8 @@
     <description>Offers matrix-based security authorization strategies (global and per-project).</description>
     <url>http://wiki.jenkins-ci.org/display/JENKINS/Matrix+Authorization+Strategy+Plugin</url>
     <properties>
-        <jenkins.version>1.609.1</jenkins.version>
-        <java.level>6</java.level>
+        <jenkins.version>2.46.1</jenkins.version>
```
oleg-nenashev: Java 8 is required only starting from 2.60.1. I would advice against doing such bump 🐛 
Java 7 for this baseline would be fine

src/main/java/hudson/security/AuthorizationProperty.java
```diff
@@ -0,0 +1,144 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package hudson.security;
```
oleg-nenashev: 🐛 please do not push new classes into packages shared with the core

daniel-beck: I actually started out with a new package but then moved stuff back. Does it make a difference? Anything that cares about this already will have a problem with this plugin.

daniel-beck:
@oleg-nenashev 

> Java 8 would require 2.60.1, no?

Ugh. Good point. I really thought it was 2.46.1. I'd really like to keep this around for the massive simplification Java 8 brings.

Could we go ahead with a review of the changes so far (assuming we'll go with 2.60.1) while we wait for September usage stats?

oleg-nenashev:
> Could we go ahead with a review of the changes so far (assuming we'll go with 2.60.1) while we wait for September usage stats?

It will require a larger discussion, but I am fine with continuing review. Will do it tomorrow. Would be great to have feedback from @andresrc regarding this baseline though.

oleg-nenashev:
Apart from the Java/Core baseline concern, the change looks good to me. Have not manually tested it though.

Regarding the package name, the question is rather about next steps in expanding the plugin functionality. In Role Strategy I had use-cases for View permission roles, individual branch permissions, better APIs, CLI commands, etc. For all of these thing we may need new classes, which may collide with other plugins and the core (we do not run PCT for the plugin on the regular basis). I would vote for having a new package like "io.jenkins.plugins.matrixauth" or whatever.

daniel-beck:
> Regarding the package name, the question is rather about next steps in expanding the plugin functionality. … I would vote for having a new package like "io.jenkins.plugins.matrixauth" or whatever.

Works for me. I moved the stuff around, kept the package name matching the groupId.

jglick:
I am fine with a 2.60.x baseline, especially considering how infrequent releases of this plugin are. If it is helpful to the codebase to use a newer baseline then we should do it.

pom.xml
jglick: Just use 2.60.3 and be done with it. That is what [most users are on anyway](http://stats.jenkins.io/pluginversions/matrix-auth.html).

src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
```diff
@@ -50,23 +45,17 @@
...
 import java.util.Set;
-import java.util.logging.Level;
-import java.util.logging.Logger;
-import jenkins.model.Jenkins;
 
 /**
  * Holds ACL for {@link ProjectMatrixAuthorizationStrategy}.
  */
-public class AuthorizationMatrixProperty extends AbstractFolderProperty<AbstractFolder<?>> {
+public class AuthorizationMatrixProperty extends AbstractFolderProperty<AbstractFolder<?>> implements AuthorizationProperty {
```
jglick: neat

src/main/java/hudson/security/ProjectMatrixAuthorizationStrategy.java
```diff
@@ -36,7 +36,6 @@
 import com.thoughtworks.xstream.converters.UnmarshallingContext;
 import com.thoughtworks.xstream.mapper.Mapper;
 import com.thoughtworks.xstream.core.JVM;
-import hudson.PluginManager;
```
jglick: gratuitous

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        @Override
+        public PermissionScope getPermissionScope() {
+            return PermissionScope.COMPUTER;
+        }
+
+        @Override
+        public AuthorizationMatrixNodeProperty newInstance(StaplerRequest req, JSONObject formData) throws FormException {
+            return createNewInstance(req, formData, false);
```
jglick: Did you check that you cannot return `null` from here?

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
```diff
@@ -0,0 +1,176 @@
...
+            return isApplicable();
+        }
+
+        @Override
+        public String getDisplayName() {
+            return "Authorization Matrix";
+        }
+
+        public FormValidation doCheckName(@AncestorInPath Computer computer, @QueryParameter String value) throws IOException, ServletException {
+            // TODO Computer needs to become a DescriptorByNameOwner for the AncestorInPath to work
```
jglick: File a core PR for it?

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
daniel-beck: No, but the custom serialization format for permissions makes me doubt that'll work.

src/main/java/hudson/security/ProjectMatrixAuthorizationStrategy.java
daniel-beck: In the first commit of four, I'd rather keep the history around than rewrite it to not clean this file up.

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
daniel-beck: https://github.com/jenkinsci/jenkins/pull/3009

daniel-beck:
@oleg-nenashev @jglick PT another L; all review comments have been addressed.

I want to keep this at Java 8, and since we lack the infra to declare a Java 8 dependency, 2.60.x. As this is (mostly) a feature, and most instances updating the plugin in the last few months are on that version or newer, I see no reason not to go with the major code simplification that Java 8 brings (which also identified that https://github.com/jenkinsci/matrix-auth-plugin/pull/9 was incomplete due to code duplication). 

Major bugfixes can always be backported to matrix-auth 1.7.x, as usual.

daniel-beck:
New test failure after 2.46 → 2.60 step due to https://github.com/jenkinsci/jenkins/pull/2790.

daniel-beck:
CC @mathias-nyman assuming you're maintaining this.

mathias-nyman:
I am not maintaining this. I merely scratched my own itch some time ago.
But kudos for the significant cleanup @daniel-beck !

PR #22, opened 2017-09-12
[JENKINS-46834] Add flexible inheritance strategies, deprecate 'blocksInheritance'

daniel-beck:
I believe this addresses the following issues in [JENKINS-46834](https://issues.jenkins-ci.org/browse/JENKINS-46834):

* [JENKINS-39873](https://issues.jenkins-ci.org/browse/JENKINS-39873), 
* [JENKINS-37904](https://issues.jenkins-ci.org/browse/JENKINS-37904), 
* [JENKINS-39409](https://issues.jenkins-ci.org/browse/JENKINS-39409), 
* [JENKINS-38356](https://issues.jenkins-ci.org/browse/JENKINS-38356), 
* possibly [JENKINS-36788](https://issues.jenkins-ci.org/browse/JENKINS-36788), 
* [JENKINS-24878](https://issues.jenkins-ci.org/browse/JENKINS-24878) (sort of),

Incidentally as part of cleanup, [JENKINS-33315](https://issues.jenkins-ci.org/browse/JENKINS-33315) is also believed to be addressed.

Supersedes https://github.com/jenkinsci/matrix-auth-plugin/pull/15.

I may rewrite history.

screen shot

@jenkinsci/code-reviewers

PR #23, opened 2017-09-12
[FIX JENKINS-20520] Save Jenkins after adding global user permission

daniel-beck:
[JENKINS-20520](https://issues.jenkins-ci.org/browse/JENKINS-20520)

@jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
+++ b/src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
@@ -1,0 +1,55 @@
+package org.jenkinsci.plugins.matrixauth;
+
+import com.gargoylesoftware.htmlunit.html.HtmlPage;
+import hudson.model.User;
+import hudson.security.ACL;
+import hudson.security.ACLContext;
+import hudson.security.GlobalMatrixAuthorizationStrategy;
+import hudson.security.HudsonPrivateSecurityRealm;
+import hudson.security.pages.SignupPage;
+import jenkins.model.Jenkins;
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.runners.model.Statement;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.RestartableJenkinsRule;
+
+public class PermissionAdderTest {
+
+    @Rule
+    public RestartableJenkinsRule r = new RestartableJenkinsRule();
+
+    @Test
+    @Issue("JENKINS-20520")
+    public void ensureSavingAfterInitialUser() {
+        r.addStep(new Statement() {
+            @Override
+            public void evaluate() throws Throwable {
+                r.j.jenkins.setSecurityRealm(new HudsonPrivateSecurityRealm(true));
+                r.j.jenkins.setAuthorizationStrategy(new GlobalMatrixAuthorizationStrategy());
+                r.j.jenkins.save();
+
+                JenkinsRule.WebClient wc = r.j.createWebClient();
+                SignupPage signup = new SignupPage(wc.goTo("signup"));
+                signup.enterUsername("alice");
+                signup.enterPassword("alice");
+                signup.enterFullName("Alice User");
+                HtmlPage success = signup.submit(r.j);
+
+                try (ACLContext _ = ACL.as(User.get("alice"))) {
+                    Assert.assertTrue(r.j.jenkins.getACL().hasPermission(Jenkins.ADMINISTER));
+                }
+            }
+        });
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
```diff
@@ -0,0 +1,55 @@
...
+public class PermissionAdderTest {
+
+    @Rule
+    public RestartableJenkinsRule r = new RestartableJenkinsRule();
+
+    @Test
+    @Issue("JENKINS-20520")
+    public void ensureSavingAfterInitialUser() {
+        r.addStep(new Statement() {
+            @Override
```
jglick: BTW once you are on 2.60.x+ you can make this a lot prettier with lambdas.

src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
```diff
@@ -0,0 +1,55 @@
...
+            public void evaluate() throws Throwable {
+                r.j.jenkins.setSecurityRealm(new HudsonPrivateSecurityRealm(true));
+                r.j.jenkins.setAuthorizationStrategy(new GlobalMatrixAuthorizationStrategy());
+                r.j.jenkins.save();
```
jglick: BTW in 2.51+ this is unnecessary: https://github.com/jenkinsci/jenkins/pull/2790

src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
```diff
@@ -0,0 +1,55 @@
+package org.jenkinsci.plugins.matrixauth;
+
+import com.gargoylesoftware.htmlunit.html.HtmlPage;
+import hudson.model.User;
+import hudson.security.ACL;
+import hudson.security.ACLContext;
+import hudson.security.GlobalMatrixAuthorizationStrategy;
+import hudson.security.HudsonPrivateSecurityRealm;
+import hudson.security.pages.SignupPage;
```
jglick: Wow, never heard of this before. Once the private realm gets moved into a plugin (?), I guess this should go into a `test-jar`.

src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
```diff
@@ -0,0 +1,55 @@
...
+
+                JenkinsRule.WebClient wc = r.j.createWebClient();
+                SignupPage signup = new SignupPage(wc.goTo("signup"));
+                signup.enterUsername("alice");
+                signup.enterPassword("alice");
+                signup.enterFullName("Alice User");
+                HtmlPage success = signup.submit(r.j);
+
+                try (ACLContext _ = ACL.as(User.get("alice"))) {
+                    Assert.assertTrue(r.j.jenkins.getACL().hasPermission(Jenkins.ADMINISTER));
```
jglick: More simply:

```java
assertTrue(r.j.jenkins.hasPermission(Jenkins.ADMINISTER));
```

More simply still, delete the `try` block and:

```java
assertTrue(r.j.jenkins.getACL().hasPermission(User.get("alice").impersonate(), Jenkins.ADMINISTER));
```

daniel-beck: Thanks, used this in https://github.com/jenkinsci/matrix-auth-plugin/pull/22 but forgot to update it here when I remembered how this works.

src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
daniel-beck: I want this to be explicit.

src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
```diff
@@ -0,0 +1,51 @@
...
+                r.j.jenkins.save();
+
+                JenkinsRule.WebClient wc = r.j.createWebClient();
+                SignupPage signup = new SignupPage(wc.goTo("signup"));
+                signup.enterUsername("alice");
+                signup.enterPassword("alice");
+                signup.enterFullName("Alice User");
+                HtmlPage success = signup.submit(r.j);
+
+                Assert.assertTrue(r.j.jenkins.getACL().hasPermission(User.get("alice").impersonate(), Jenkins.ADMINISTER));
```
jglick, line 41: Failing in https://github.com/jenkinsci/bom/pull/359 for reasons I am still struggling to understand: `User.get("alice", false, Collections.emptyMap())` returns null. Only when `jth.jenkins-war.path` is set to the megawar.

PR #24, opened 2017-09-13
[JENKINS-46832] Prevent initial misconfiguration by making sure there's an admin

daniel-beck:
This is a pretty crude approach to attempt to prevent the problem, but I'm not sure with the current form there can be a better one.

This addresses part of [JENKINS-9774](https://issues.jenkins-ci.org/browse/JENKINS-9774) by preventing accidentally empty global permissions overall (but not the parse problem -- that's handled by https://github.com/jenkinsci/matrix-auth-plugin/pull/25), and part of [JENKINS-10871](https://issues.jenkins-ci.org/browse/JENKINS-10871) by ensuring there's an admin that can log in (or, in the worst case, anonymous).

While the UX of [JENKINS-28668](https://issues.jenkins-ci.org/browse/JENKINS-28668) remains terrible, this change will prevent such accidental lockouts as well.

The current approach will result in anonymous users being granted admin permissions in many cases, instead of just saving a broken configuration resulting in lock-out. This should still be preferable to just locking out users, and should be pretty obvious when browsing Jenkins as well.

@jenkinsci/code-reviewers

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.33</version>
+        <version>2.29</version>
     </parent>
     <artifactId>matrix-auth</artifactId>
     <version>1.8-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


daniel-beck:
Merging to unblock writing tests in other PRs (as this PR adds the test class), but reviews are still appreciated.

PR #26, opened 2017-09-14
[FIX JENKINS-14563] Show full name of user instead of user name

daniel-beck:
For users that have a "full name" different from their user ID, this will show the full name, with the user ID available as tool tip.

screen shot

@jenkinsci/code-reviewers mostly to get opinions on usefulness / possible unintended side effects

### DIFF
--- a/src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
+++ b/src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
@@ -30,6 +30,7 @@ import com.thoughtworks.xstream.io.HierarchicalStreamReader;
 import com.thoughtworks.xstream.io.HierarchicalStreamWriter;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.PluginManager;
+import hudson.Util;
 import hudson.diagnosis.OldDataMonitor;
 import hudson.model.Descriptor;
 import jenkins.model.IdStrategy;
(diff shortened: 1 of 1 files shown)


src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
```diff
@@ -473,12 +473,16 @@ public FormValidation doCheckName_(@Nonnull String value, @Nonnull AccessControl
...
             if(v.equals("authenticated"))
                 // system reserved group
-                return FormValidation.respond(Kind.OK, makeImg("user.png", "Group", false) +ev);
+                return FormValidation.respond(Kind.OK, makeImg("user.png", ev, "Group", false));
 
             try {
                 try {
                     sr.loadUserByUsername(v);
-                    return FormValidation.respond(Kind.OK, makeImg("person.png", "User", false)+ev);
+                    User u = User.get(v);
```
oleg-nenashev: This call will make Jenkins to create a user even if a bogus security realm didn't load/report it properly. It is not a big deal, but I would recommend to get user without creation and use `if (u == null || ev.equals(u.getFullName()))` in the next clause. 🐜 .

src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
```diff
@@ -473,12 +473,16 @@ public FormValidation doCheckName_(@Nonnull String value, @Nonnull AccessControl
...
+                    if (ev.equals(u.getFullName())) {
+                        return FormValidation.respond(Kind.OK, makeImg("person.png", ev, "User", false));
+                    }
+                    return FormValidation.respond(Kind.OK, makeImg("person.png", u.getFullName(), "User " + ev, false));
```
oleg-nenashev: This name is unlimited. I can imagine cases where the REALLY long name breaks the UI. The worst thing I have seem on my instance (not the one you would think about, another one) was somewhat more than 64 symbols because it was corrupting our user database in SCM

src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
daniel-beck: It's an ephemeral user though. And only ever created if it's configured (or about to be configured) in the security realm. Doesn't seem like a big deal to me.

daniel-beck:
@oleg-nenashev Thanks, fixed:

screen shot

PR #27, opened 2017-09-14
[FIX JENKINS-5277] Ensure that creators have read and configure access

daniel-beck:
@jenkinsci/code-reviewers


daniel-beck:
Yay tests: Turns out I had messed up in https://github.com/jenkinsci/matrix-auth-plugin/pull/18 and didn't actually make use of the node property in the security realm. Oops.

PR #14, opened 2016-12-20
[JENKINS-34616] add @DataBoundConstructor, README.md with sample pipeline

mknezic:
add @DataBoundConstructor, README.md with sample pipeline, should help with https://issues.jenkins-ci.org/browse/JENKINS-34616

PR #32, opened 2017-09-16
[JENKINS-34616] Add symbols, getter, and constructor for pipeline

daniel-beck:
Supersedes https://github.com/jenkinsci/matrix-auth-plugin/pull/14

[JENKINS-34616](https://issues.jenkins-ci.org/browse/JENKINS-34616)

- ~~I don't like the fully qualified permission IDs – ideally, they would be shortened to "Class.Field" if unambiguous.~~ (JENKINS-17200)
- ~~Needs usage docs~~ Optional, sort of, snippet gen works after all.
- [x] Needs autotests, manual testing etc.

----

Snippet generator:

> screen shot

Result (formatted for readability):

```groovy
properties([
    authorizationMatrix(
        inheritanceStrategy: nonInheriting(), 
        permissions: [
            'hudson.model.Item.Read:authenticated', 
            'hudson.model.Run.Delete:danielbeck', 
            'hudson.model.Item.Build:danielbeck', 
            'hudson.model.Item.Configure:danielbeck']
            ),
    pipelineTriggers([])])
```

@jenkinsci/code-reviewers assemble!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,13 @@
             <version>2.10</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <!-- requireUpperBoundDeps -->
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>structs</artifactId>
+            <version>1.6</version>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <repositories>
         <repository>
--- a/pom.xml
+++ b/pom.xml
@@ -40,6 +40,19 @@
             <version>2.1.11</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins.workflow</groupId>
+            <artifactId>workflow-cps</artifactId>
+            <version>2.30</version>
+            <classifier>tests</classifier>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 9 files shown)


daniel-beck:
> I don't like the fully qualified permission IDs – ideally, they would be shortened to "Class.Field" if unambiguous.

Doesn't work with `SCM.Tag`, as `Permission#getAll()` doesn't include it (likely [JENKINS-17200](https://issues.jenkins-ci.org/browse/JENKINS-17200)). So this would be too brittle right now to support it, but could always be added later.

jglick:
I see no problem with the current format. Better to avoid ambiguity.

BTW `pipelineTriggers([])` is removed by https://github.com/jenkinsci/workflow-job-plugin/pull/42 which is waiting for a fix to https://github.com/jenkinsci/workflow-job-plugin/pull/41.

daniel-beck:
Not sure how to implement the restriction of eligible permissions here. I suppose I'll wait for https://github.com/jenkinsci/matrix-auth-plugin/pull/33 first to not introduce a greater mess. Filed [JENKINS-46944](https://issues.jenkins-ci.org/browse/JENKINS-46944)

daniel-beck:
I'm not super happy with the system wide logging without providing a very useful context for this. Don't have a much better idea right now.

PR #33, opened 2017-09-16
Further cleanup addressing more copied code

daniel-beck:
Extract an interface with default method implementations common to the strategy and the properties (as opposed to the existing one that only did code common to properties), an interface for their descriptors, and a new abstract converter.

Untested.

### DIFF
--- a/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
+++ b/src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
@@ -98,7 +98,7 @@ public class AuthorizationMatrixProperty extends AbstractFolderProperty<Abstract
 
     @Restricted(NoExternalUse.class)
     public Set<String> getGroups() {
-        return sids;
+        return new HashSet<>(sids);
     }
 
     /**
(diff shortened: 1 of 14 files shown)


bpedersen2:
Fixes JENKINS-46923

src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
```diff
@@ -96,6 +96,7 @@ public AuthorizationMatrixProperty(Map<Permission,? extends Set<String>> granted
             this.grantedPermissions.put(e.getKey(),new HashSet<>(e.getValue()));
     }
 
+    @Restricted(NoExternalUse.class)
```
oleg-nenashev: Why? I understand that nobody is expected to write to this method. `Collections.unmodifyableSet()` would be reasonable, but I do not see issues with having getter in API

daniel-beck: This only exists to build up the potential groups for http://javadoc.jenkins.io/hudson/security/AuthorizationStrategy.html#getGroups-- and is not intended for any other use.

oleg-nenashev: Collections.unmodifyableSet() would be useful anyway

daniel-beck: I went with a copy of the set instead, consistent with some other getters that return a newly built collection.

daniel-beck:
Merging to get a beta-2 out, further feedback is always appreciated @oleg-nenashev

PR #40, opened 2017-10-21
[JENKINS-47555] Fix test on 2.73.2+

daniel-beck:
[JENKINS-47555](https://issues.jenkins-ci.org/browse/JENKINS-47555) by @andresrc 

CC @jglick Perhaps `CommandLauncher` isn't a great default for `JenkinsRule` anymore?


jglick:
> Perhaps `CommandLauncher` isn't a great default for `JenkinsRule` anymore?

It is not used any more. You just need a newer parent, thus JTH.

PR #42, opened 2017-11-09
Do not perform any checks for System user

Jimilian:
According the Jenkins Core docs in case of SYSTEM user we should assume
that it can do whatever he wants.

p.s. The original problem comes from `getPlugin` method that is already fixed in Core ( https://github.com/jenkinsci/jenkins/pull/3109)

screen shot 2017-11-09 at 17 32 51


Jimilian:
@daniel-beck, can you take a look please?
I see that CI failed, but locally everything passed and I think it's some azure related problem.

daniel-beck:
@Jimilian Thanks 👍

PR #45, opened 2018-01-30
[JENKINS-49199] added @DataBoundConstructor

daspilker:
[JENKINS-49199](https://issues.jenkins-ci.org/browse/JENKINS-49199)

PR #46, opened 2018-02-08
Add configuration-as-code support with extra contructor

MadsNielsen:
We do not need to do major work to support c-as-c. Adding an extra constructor enables this.

### DIFF
--- a/src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
+++ b/src/main/java/hudson/security/GlobalMatrixAuthorizationStrategy.java
@@ -72,7 +72,7 @@ public class GlobalMatrixAuthorizationStrategy extends AuthorizationStrategy imp
     }
 
     /**
-     * Added because of <a href="https://github.com/jenkinsci/configuration-as-code-plugin"/>Configuration as Code plugin</a>
+     * Added because of <a href="https://github.com/jenkinsci/configuration-as-code-plugin">Configuration as Code plugin</a>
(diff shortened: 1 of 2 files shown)


daniel-beck:
This needs better justification. Is it because the map has no setter?

Also, needs Javadoc and `@since TODO`.

MadsNielsen:
@daniel-beck Yep, i need to set the `grantedPermissions` map. I'm open to suggestions on how to do it otherwise. I can easilly add the javadoc etc.

MadsNielsen:
@daniel-beck And i could probably do the same using reflection in configuration-as-code plugin but having a simple constructor to generate the Authentication section option seemed to be a better solution.

daniel-beck:
@MadsNielsen Not opposed, just the reason wasn't clear, and no explanation given.

MadsNielsen:
@daniel-beck Just awright, i actually did forget to add a justification on the commit. So i've updated my pull request with a bit of JavaDoc and some `@since`'s to be filled out.

daniel-beck:
@MadsNielsen How does this relate to https://github.com/jenkinsci/matrix-auth-plugin/pull/45? Is the other also relevant for you?

MadsNielsen:
@daniel-beck It's probably a better solution than the one i proposed, but yes, essentially we're trying to solve the same problem. I did not know that the form you submit could be condensed to a simple list of strings, as you have a group name for each set of granted permissions.

daniel-beck:
Making this nicer in https://github.com/jenkinsci/matrix-auth-plugin/pull/46

PR #48, opened 2018-05-16
[JENKINS-56805] Support Configuration as Code

ndeloof:
Move configuration-as-code support in matrix-auth plugin

Configuration-as-Code [JEP-201](https://github.com/jenkinsci/jep/tree/master/jep/201) allow to configure a full jenkins master from a plain text definition. We'd like this feature hosted by the implementor.

see jenkinsci/configuration-as-code-plugin#197


daniel-beck:
@ndeloof Sorry for the late response. This PR doesn't build (anymore), could you please look into that?

The test and sample config look reasonable, but please fix up license headers & Javadoc.

There's also #46 and I need your and @MadsNielsen's help to figure out which is the better approach, as I'm not sure what's going on exactly -- are they alternatives, if so, what are the pros/cons? Or do they complement each other?

Is there some synergy for CasC and Job DSL (#45) to be had, or are they too dissimilar for that?

ndeloof:
I'm not sure about #46 intent, I haven't seen the "JCasC side" of this PR. But if we get Configurator hosted in plugin this probably would make more sense.

About #45 I don't know Job DSL enough to tell you the exact impact but I guess it relies on comparable mechanisms, just focussing on job configuration. We both need the same level of "cleanup" in jenkins plugins :)

Will update this PR once 0.10-alpha has been released.

daniel-beck:
@ndeloof Thanks for the explanation. Please ping me out of band if I miss the PR update.

ndeloof:
updated with some changes :
custom converter now uses JCasC API (vs fully reimplementing `configure`) and uses a flatten permission list model : 
```yaml
    projectMatrix:
      grantedPermissions:
        - "Overall/Read:anonymous"
        - "Overall/Administer:authenticated"
```

waiting for 0.10-alpha release

daniel-beck:
Still failing. Please specify the snapshot dependency using a timestamp for the artifact you uploaded.

ndeloof:
I don't understand this access modifier failure :-\

daniel-beck:
@ndeloof Most were due to outdated parent POM.

daniel-beck:
Test failures since the configuration-as-code snapshot still includes the classes now moved here?

ndeloof:
cyclic dependency, as I should have anticipated. Need to remove dependency from JCasC

daniel-beck:
What's the plan to release this + that? Ideally I'd like to give this a spin without time pressure while a release of CasC is out that doesn't support matrix-auth anymore, or are your early adopters unusually accepting of breakage?

casz:
As an early adopter I say early breakage is expected 👍 It is still alpha for that reason

ndeloof:
@daniel-beck I'd expect custom configurator removed for JCasC in next alpha release (some weeks at most).
We broke things few times already, so no pressure :)

daniel-beck:
I don't have a better label meaning "do not merge" right now 😉

ndeloof:
oh ok, no worries then.

casz:
@daniel-beck would be create to get #51 and #50 merged so we could test the incrementals version in JCasc 👍

x10an14:
@daniel-beck Hi! Just wanted to let you know that as a user of JCasC, I (and I imagine others) are eagerly awaiting the day when you get an opportunity to look at this PR =)

daniel-beck:
@ndeloof and perhaps @casz Could you please clarify whether this should be published?

ndeloof:
@daniel-beck no, this is for pre-review. We wait for Release Candidates to ensure API is stable.

ndeloof:
@daniel-beck updated after JCasC 1.0 release, can be merged now if you're ok with the implementation

casz:
@ndeloof how will this work with already installed jcasc-support plugin ?

ndeloof:
@casz support implementation will be picked-up first (ordinal=1). Once matrix-auth-plugin has been release we can declare optional dependency on this new version and remove code from support module, so we enforce support-plugin users to get the fresh new version hosted by matrix auth if they upgrade. This will prevent them to upgrade support and suddenly miss this configurator.

daniel-beck:
JW next week so this will probably take a week or two at least.

casz:
What are the chances of getting a release? @daniel-beck @ndeloof 😅 Currently the jcasc-support plugin does not support export of permission but they are fixed here and mentioned in the linked issue above

tomasquith:
@ndeloof @daniel-beck Is there anything that needs doing in order to release this? It'd be lovely to move to config as code, but this is the only issue stopping us

casz:
@daniel-beck would be nice with a release 🙌

daniel-beck:
Currently looking into the history of this integration. For my own reference:

Current location:
https://github.com/jenkinsci/configuration-as-code-plugin/tree/master/support/src/main/java/io/jenkins/plugins/casc/support/matrixauth

Previous location:
https://github.com/jenkinsci/configuration-as-code-plugin/tree/2eee696c8862f714ca87f13343b47c2d5dc9ef61/support/src/main/java/org/jenkinsci/plugins/casc/support/matrixauth

Previous previous location:
https://github.com/jenkinsci/configuration-as-code-plugin/tree/3ea44fa10ef129a073b473591ac03dc520dd15a7/plugin/src/main/java/org/jenkinsci/plugins/casc/integrations/globalmatrixauth

Previous previous previous location:
https://github.com/jenkinsci/configuration-as-code-plugin/tree/157b31c0462f697e11e80d6b0ff7644cbfd331b3/src/main/java/org/jenkinsci/plugins/casc/integrations/globalmatrixauth

Previous previous previous previous location:
https://github.com/jenkinsci/configuration-as-code-plugin/tree/7f1aaccf9ebc072cc361d56821145b76aeea3daa/src/main/java/org/jenkinsci/plugins/casc/integrations

ndeloof:
@daniel-beck don't look too much into history, API has changed a bit during alpha-beta-rc phases.

ndeloof:
please also note that here we just propose our custom implementation for adoption, but you can also evolve plugin codebase to just expose the required javabean accessors: 

AuthorizationContainer#setXXX(Collection<String>)
AuthorizationContainer#getXXX() : Collection<String>

(XXX to be define with adequate nouns to avoid collision with existing getGrantedPermissions relying on Map)

src/test/java/org/jenkinsci/plugins/matrixauth/ConfigurationAsCodeTest.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Created by mads on 2/22/18.
+ */
+public class ConfigurationAsCodeTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    public void should_configure_permissions() throws Exception {
```
timja: Missing a test case for exporting the config here:
https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/docs/PLUGINS.md#model-export-test

The current released support for matrix auth plugin and export with configuration is broken.
It successfully exports but won't accept the exported version and fails with illegalstateexception

ndeloof: would be nice indeed, even we don't have a yaml assertion library to test actual export. Also, export being highly experimental and missing support from many components, including jenkins-core, I won't consider this a blocker.

timja: not a blocker but in general when export doesn't stacktrace it does work, unfortunately in what's released currently it doesn't :(

timja: @ndeloof I've just tested this branch,
Is it expected that export would export it in this form?
```
  authorizationStrategy:
    globalMatrix:
      grantedPermissions:
      - "Permission[class hudson.model.View,Delete]:authenticated"
      - "Permission[class hudson.model.Computer,Connect]:authenticated"
      - "Permission[class hudson.model.Run,Delete]:authenticated"
      - "Permission[class hudson.model.Computer,Create]:authenticated"
      - "Permission[class hudson.model.View,Configure]:authenticated"
      - "Permission[class hudson.model.Computer,Build]:authenticated"
      - "Permission[interface hudson.model.Item,Configure]:authenticated"
      - "Permission[class hudson.model.Hudson,Administer]:authenticated"
      - "Permission[class hudson.model.Hudson,Administer]:300e771f-856c-45cc-b899-40d78281e9c1"
```

ndeloof: No. We need to provide a custom converter for hudson.security.Permission. Initial adapter implementation has been designed for this specific plugin (before export was implemented) but this is a more generic need.

timja: to be done in this PR or separate? I've had a bit of a look at the code but doesn't look easy from what I can see, I assume its a change to `ConifigurationAsCode#toYaml`

timja:
@daniel-beck is there anything blocking this PR?
the export support is a new feature that JCasC would need for this configurator

daniel-beck:
@timja Time constraints, mostly. I want to get this done myself, but it's competing for my spare time.

casz:
The API has had a few changes and we have an PR that changes the API further, if you'd like I could get commit access to fix up the pr

jonbrohauge:
Linking: https://github.com/jenkinsci/configuration-as-code-plugin/issues/809

daniel-beck:
I'm currently trying to add support for the configuration of matrix permissions on agents. They, as well as per-job permissions (which are still out of scope for CasC itself IIUC), have a field `inheritanceStrategy`. It seems that https://github.com/jenkinsci/configuration-as-code-plugin/blob/782f87e8cd81d9d49a3b32b3a3028eca41b74b8e/plugin/src/main/java/io/jenkins/plugins/casc/impl/configurators/DataBoundConfigurator.java#L258 refuses to consider a Describable with no attributes a valid object, returning `null`, and ultimately failing the export.

There are several valid use cases for something like this off the top of my head: The inheritance strategy here (see https://github.com/jenkinsci/matrix-auth-plugin/tree/master/src/main/java/org/jenkinsci/plugins/matrixauth/inheritance ) allows Describables to control behavior through their identity alone, no options needed. Similarly, there's a plausible argument that MarkupFormatters in core don't _need_ options either.

Is this a bug in CasC? Or am I missing something?

casz:
How would you expect this to look?

Based on this.
```yaml
jenkins:
  nodes:
    - permanent:
        labelString: "linux docker test"
        mode: NORMAL
        name: "utility-node"
        remoteFS: "/home/user1"
        launcher:
          jnlp:
```

You could argue an empty `InheritParentStrategy:` or `InheritGlobalStrategy:` is valid. 
You could write your configurator differently instead of using descriableAttribute.
So that you would have:
`strategy: "inheritParent"`

I think the reason for removing empty but I if it makes sense to change it I am open.

daniel-beck:
@casz I think the following would make sense (as usual with the disclaimer I'm not a user of CasC, plus haven't checked whether this is legal YAML):

````
  nodes:
  - dumb:
      name: "agent"
      nodeProperties:
      - authorizationMatrix:
          inheritanceStrategy: nonInheriting
          permissions:
          - "Agent/Build:anonymous"
````

Add an argument and the format would be how it is today. Assuming this can be round-tripped reliably, why not?

There just needs to be a way to gracefully scale down to 0 attributes. Any extension point may end up with that.

In fact, on a default install of Jenkins, as expected in a previous comment, the default "Plain text" markup formatter seems to fail an export for the same reason. "Safe HTML" works but only because it has a (really non-essential) attribute. An agent's default retention strategy ("keep online as much as possible") also seems to fail to export because it has no further options, with one that has further attributes, it works. This indicates to me that the way things work today is broken.

File: src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java [hunk 2]
```diff
@@ -100,6 +104,14 @@ public AuthorizationMatrixProperty(Map<Permission,? extends Set<String>> granted
             this.grantedPermissions.put(e.getKey(),new HashSet<>(e.getValue()));
     }
 
+    @DataBoundConstructor // JENKINS-49199: Used for job-dsl
+    @Restricted(NoExternalUse.class)
+    public AuthorizationMatrixProperty(List<String> permissions) {
```
casz: This could be a databound setter, not to tie you down to using a constructor and deprecating and all.

daniel-beck: I would not consider permissions optional (which is, at least in Pipeline, the defining difference between what goes into a constructor and what is just a setter), and since it's `@Restricted` I don't need to do the `@Deprecated` dance in the future either.

src/test/resources/org/jenkinsci/plugins/matrixauth/Configuration-as-Code.yml
```diff
@@ -0,0 +1,7 @@
+jenkins:
```
casz: Hmmm, seems obslete :confused:

src/test/java/org/jenkinsci/plugins/matrixauth/integrations/ConfigurationAsCodeTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jenkinsci.plugins.matrixauth.inheritance.InheritGlobalStrategy;
+import org.jenkinsci.plugins.matrixauth.inheritance.NonInheritingStrategy;
+import org.junit.Rule;
+import org.junit.Test;
+import io.jenkins.plugins.casc.misc.ConfiguredWithCode;
+import io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule;
+
+import static org.junit.Assert.*;
+
+public class ConfigurationAsCodeTest {
```
casz: Two tests with the same name :confounded:

casz:
@daniel-beck I could really use an incremental :laughing:

src/main/java/org/jenkinsci/plugins/matrixauth/integrations/PermissionFinder.java
```diff
@@ -0,0 +1,87 @@
...
+        Matcher m = PERMISSION_PATTERN.matcher(id);
+        if(m.matches()) {
+            String owner = m.group(1);
+            String name = m.group(2);
+            for(PermissionGroup pg : pgs) {
+                if(pg.owner.equals(Permission.class)) {
+                    continue;
+                }
+
+                //TODO: Support PermissionGroupIDs (JENKINS-51598)
```
casz: Remove todo :wink:

src/main/java/org/jenkinsci/plugins/matrixauth/integrations/casc/MatrixAuthorizationStrategyConfigurator.java
```diff
@@ -0,0 +1,60 @@
...
+        );
+    }
+
+    /**
+     * Extract container's permissions as a List of "PERMISSION:sid"
+     */
+    public static Collection<String> getGrantedPermissions(AuthorizationContainer container) {
+        return container.getGrantedPermissions().entrySet().stream()
+                .flatMap( e -> e.getValue().stream()
+                        .map(v -> e.getKey().group.title.toString(Locale.US)+"/"+e.getKey().name+":"+v))
```
casz: Same as above use `group.getId()`

```suggestion
                        .map(v -> e.getKey().group.getId()+"/"+e.getKey().name+":"+v))
```

src/main/java/org/jenkinsci/plugins/matrixauth/integrations/casc/MatrixAuthorizationStrategyConfigurator.java
```diff
@@ -0,0 +1,60 @@
...
+    public Class<?> getImplementedAPI() {
+        return AuthorizationStrategy.class;
+    }
+
+
+    @Override
+    @Nonnull
+    public Set<Attribute<T, ?>> describe() {
+        return Collections.singleton(
+                new MultivaluedAttribute<T, String>("grantedPermissions", String.class)
```
daniel-beck:
```suggestion
                new MultivaluedAttribute<T, String>("permissions", String.class)
```

Already done for the `NodeProperty`, just not here.

casz: this will break those that currently use it :cry: but you are of course allowed, just make it clear :)

src/main/java/org/jenkinsci/plugins/matrixauth/integrations/casc/MatrixAuthorizationStrategyConfigurator.java
```diff
@@ -57,4 +65,14 @@ public static void setGrantedPermissions(AuthorizationContainer container, Colle
             container.add(permission, p.substring(i+1));
         });
     }
+
+    /**
+     * Like {@link #setGrantedPermissions(AuthorizationContainer, Collection)} but logs a deprecation warning
+     */
+    public static void setGrantedPermissionsDeprecated(AuthorizationContainer container, Collection<String> permissions) {
+        LOGGER.log(Level.WARNING, "Loading deprecated attribute 'assignedPermissions' for instance of '" + container.getClass().getName() +"'. Use 'permissions' instead.");
```
casz:
```suggestion
        LOGGER.log(Level.WARNING, "Loading deprecated attribute 'grantedPermissions' for instance of '" + container.getClass().getName() +"'. Use 'permissions' instead.");

```

src/main/java/org/jenkinsci/plugins/matrixauth/integrations/casc/MatrixAuthorizationStrategyConfigurator.java
```diff
@@ -30,11 +33,16 @@
...
     @Nonnull
     public Set<Attribute<T, ?>> describe() {
-        return Collections.singleton(
-                new MultivaluedAttribute<T, String>("grantedPermissions", String.class)
+        return new HashSet<>(Arrays.asList(
+                new MultivaluedAttribute<T, String>("permissions", String.class)
                         .getter(MatrixAuthorizationStrategyConfigurator::getGrantedPermissions)
-                        .setter(MatrixAuthorizationStrategyConfigurator::setGrantedPermissions)
-        );
+                        .setter(MatrixAuthorizationStrategyConfigurator::setGrantedPermissions),
```
casz: Should the get and set not be renamed too?

daniel-beck: Unsure. Ultimately they end up setting the items in the `grantedPermissions` field.

src/test/resources/org/jenkinsci/plugins/matrixauth/integrations/configuration-as-code.yml
```diff
@@ -0,0 +1,73 @@
+jenkins:
+  authorizationStrategy:
+    projectMatrix:
```
timja: may have missed it but is there a test needed for globalMatrix?

src/test/java/org/jenkinsci/plugins/matrixauth/integrations/ConfigurationAsCodeTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.jenkinsci.plugins.matrixauth.inheritance.NonInheritingStrategy;
+import org.junit.Rule;
+import org.junit.Test;
+import io.jenkins.plugins.casc.misc.ConfiguredWithCode;
+import io.jenkins.plugins.casc.misc.JenkinsConfiguredWithCodeRule;
+import org.jvnet.hudson.test.LoggerRule;
+
+import static org.junit.Assert.*;
+
+public class ConfigurationAsCodeTest {
```
timja: export test?

daniel-beck: Perhaps if I decide I have way too much time. Until then I'll consider it something to be implemented once

> Screen Shot


exists.

src/test/resources/org/jenkinsci/plugins/matrixauth/integrations/configuration-as-code.yml
daniel-beck: Would consider that redundant due to how inheritance is set up.

src/test/java/org/jenkinsci/plugins/matrixauth/integrations/ConfigurationAsCodeTest.java
timja: https://github.com/jenkinsci/configuration-as-code-plugin/pull/851

src/test/java/org/jenkinsci/plugins/matrixauth/integrations/casc/ExportTest.java
```diff
@@ -0,0 +1,60 @@
...
+            AuthorizationStrategy authorizationStrategy = j.jenkins.getAuthorizationStrategy();
+            Configurator c = context.lookupOrFail(ProjectMatrixAuthorizationStrategy.class);
+
+            @SuppressWarnings("unchecked")
+            CNode node = c.describe(authorizationStrategy, context);
+            assertNotNull(node);
+            Mapping mapping = node.asMapping();
+
+            List<CNode> permissions = mapping.get("permissions").asSequence();
+            assertEquals("list size", 18, permissions.size());
```
timja: maybe assert the list contains one of the expected string?
e.g.
```
"Job/Build:authenticated"
```

timja:
No incrementals due to:
```
Skipping deployment as no artifacts were found with the expected path, typically due to a PR merge build not up to date with its base branch
```
I assume master needs merging in for an incrementals build?

daniel-beck:
@timja I did that, deployment is https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/matrix-auth/2.4-rc247.5d90c4bd1f21/

timja:
Nice, anything left to do on this?

daniel-beck:
Ideally more reviews confirming I didn't take a wrong turn somewhere, since I'm not a user of any of this myself.

src/test/java/org/jenkinsci/plugins/matrixauth/integrations/casc/ExportTest.java
```diff
@@ -0,0 +1,72 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class ExportTest {
+
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
+
+    @Test
+    @LocalData
```
casz: Hmm feels odd not to use JCasC to import the same input data.

daniel-beck: Going from non-CasC configuration to CasC export is a more expressive test case. Otherwise I need to "know" that CasC takes no shortcuts from between import and export (e.g. just echo back whatever it "imported" and never even applied to the running config).

PR #50, opened 2018-07-13
Incrementals

casz:
cc @daniel-beck @oleg-nenashev


casz:
#48 should fix the CI: Access modifier checks failed

daniel-beck:
Subsumed by https://github.com/jenkinsci/matrix-auth-plugin/pull/53

PR #53, opened 2018-10-11
Add dependency on non-tests classifier version of workflow-cps and incrementalify

dwnusbaum:
Although the plugin has an explicit dependency on the `<classifier>tests</classifier>` version of `workflow-cps`, the regular test-scope dependency is pulled in transitively through `workflow-multibranch`. 

This works fine for now since the 2 current versions seem to be compatible, but in the PCT, there can be failures because `SnippitizerTester` (which comes from the `<classifier>tests</classifier>` version of `workflow-cps`) from `workflow-cps` 2.56+ is not compatible with regular `workflow-cps` 2.55-, and the PCT isn't smart enough to update the non-classifier version unless it is specified explicitly here (it updates the `workflow-multibranch` version, but that is not enough).

I also incrementalified the plugin while I was here and fixed the access-modifier issue.

### DIFF
--- a/.mvn/extensions.xml
+++ b/.mvn/extensions.xml
@@ -1,0 +1,7 @@
+<extensions xmlns="http://maven.apache.org/EXTENSIONS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/EXTENSIONS/1.0.0 http://maven.apache.org/xsd/core-extensions-1.0.0.xsd">
+  <extension>
+    <groupId>io.jenkins.tools.incrementals</groupId>
+    <artifactId>git-changelist-maven-extension</artifactId>
+    <version>1.0-beta-7</version>
+  </extension>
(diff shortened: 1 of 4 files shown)


dwnusbaum:
Sorry @casz I didn't see your PRs before submitting this or I would have based it off of them. This should subsume #50 and #51.

jvz:
Ok now I could really use the incrementals on this repo so I can continue work on [JENKINS-21336](https://issues.jenkins-ci.org/browse/JENKINS-21336).

casz:
https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/matrix-auth/2.4-rc198.ac6c813a7ee8/ does that help? at least this PR is incremental 🤭

jvz:
I suppose I could try forking from this PR. Good point!

daniel-beck:
Thanks @dwnusbaum and @casz for these changes, and I apologize for the delay. I've been very busy recently, so this plugin got a bit neglected.

PR #55, opened 2019-01-04
[JENKINS-55397] Updating AuthorizationContainerDescriptorTest for new…

fcojfernandez:
See [JENKINS-55397](https://issues.jenkins-ci.org/browse/JENKINS-55397)

Using newer core versions (starting at 2.120),` AuthorizationContainerDescriptorTest#testImpliedNotes` is falling in the PCT. The error is caused by [this change](https://github.com/jenkinsci/jenkins/commit/081e400a5b618b035fa9c350e3bf4341d66e2476). From then, the `Item.CANCEL` permission is not imply by `BUILD` but by `Permission.UPDATE`, so its description changes. Having made a research, it seems that only the description text is affected, so the plugin is still compatible with Jenkins-core and only the test should be updated.

@reviewbybees 
@daniel-beck as maintainer
@raul-arabaolaza who found the issue

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptorTest.java
+++ b/src/test/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptorTest.java
@@ -18,9 +18,6 @@ import java.util.Locale;
 
 public class AuthorizationContainerDescriptorTest {
 
-    @Rule
-    public JenkinsRule r = new JenkinsRule(); // Needed to check the jenkins version
-
     private Permission TEST_PERMISSION = new Permission(Item.PERMISSIONS, "Test", new Localizable(new ResourceBundleHolder(AuthorizationContainerDescriptorTest.class), "Test"), Item.BUILD, PermissionScope.ITEM);
 
     @Test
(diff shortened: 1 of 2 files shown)


daniel-beck:
Seems reasonable as a stop gap, but would really like there to be a test even on 2.120+ for permissions with the added note. Perhaps a permission defined for the test only? Haven't had the time to look into it yet, unfortunately.

fcojfernandez:
I think it's worth to keep the test for the Item.CANCEL Permission but I've added a new one only for testing the added note. Let me know your thoughts about my last commit, @daniel-beck

daniel-beck:
@fcojfernandez PTAL

src/test/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptorTest.java
```diff
@@ -21,7 +21,7 @@
     @Rule
     public JenkinsRule r = new JenkinsRule(); // Needed to check the jenkins version
 
-    private Permission FAKE = new Permission(Item.PERMISSIONS, "Fake", new TestLocalizable(), Item.BUILD, PermissionScope.ITEM);
+    private Permission TEST_PERMISSION = new Permission(Item.PERMISSIONS, "Test", new Localizable(new ResourceBundleHolder(AuthorizationContainerDescriptorTest.class), "Test"), Item.BUILD, PermissionScope.ITEM);
```
fcojfernandez: I tried to do this but I didn't find the way. I saw [here](https://wiki.jenkins.io/display/JENKINS/Internationalization), that _Jenkins uses localizer to generate classes for accessing message resources in a type-safe way. For each src/main/resources/**/Messages.properties, a Messages class is generated. See the referenced page for how this class looks like. If your IDE fails to find these classes, manually add "target/generated-sources/localizer" directory to your source root._, so I thought that a properties won't be loaded as test resource. I had a look at the [maven-localizer-plugin](https://github.com/kohsuke/localizer) that is configured in parent-pom as well and it make me think my first research was ok.
Good to know. Thanks!

src/test/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptorTest.java
```diff
@@ -37,14 +34,8 @@ public void testImpliedNotes() {
             Assert.assertFalse(description.contains(Messages.GlobalMatrixAuthorizationStrategy_PermissionImpliedBy(Jenkins.PERMISSIONS.title.toString(), Jenkins.ADMINISTER.name)));
         }
 
-        { // Item.CANCEL is implied by Item.BUILD until 2.119. From 2.120 is implied by Permission.UPDATE, so the description changes.
```
fcojfernandez: You have more context about if it's worth to maintain the test for Item.CANCEL, so I'm happy to remove it.

daniel-beck: Simply chosen as an example of its category of being implied by a non-Overall/Administer permission.

src/test/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptorTest.java
daniel-beck: Hacky approach, obviously: We're bypassing the code generation here.

PR #59, opened 2019-04-24
Allow using internal permission identifiers in CasC

daniel-beck:
@timja @casz Is this a use case that makes sense?


src/main/java/org/jenkinsci/plugins/matrixauth/integrations/casc/MatrixAuthorizationStrategyConfigurator.java
```diff
@@ -51,7 +51,7 @@
         return container.getGrantedPermissions().entrySet().stream()
                 .flatMap( e -> e.getValue().stream()
                         .map(v -> e.getKey().group.getId() + "/" + e.getKey().name + ":" + v))
-                .collect(Collectors.toList());
+                .sorted().collect(Collectors.toList());
```
daniel-beck: I don't know how this snuck in here – makes sense for export independently.

casz:
Why expose users to Java package naming mainly when Hudson still lurks around :laughing:

PR #60, opened 2019-04-25
[JENKINS-46914] [JENKINS-47885] [JENKINS-54568] JavaScript improvements

daniel-beck:
* [JENKINS-54568](https://issues.jenkins-ci.org/browse/JENKINS-54568): Make `authorizationMatrix` work in declarative snippet generator
* [JENKINS-46914](https://issues.jenkins-ci.org/browse/JENKINS-46914): Better indicate implied permissions in the checkbox grid by disabling implied permission checkboxes
* [JENKINS-47885](https://issues.jenkins-ci.org/browse/JENKINS-47885): Make node property work in Kubernetes (and old versions of Docker) plugin templates

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
@@ -179,7 +179,7 @@ public class AuthorizationMatrixNodeProperty extends NodeProperty<Node> implemen
 
         @Override
         public boolean isApplicable(Class<? extends Node> node) {
-            return Node.class.isAssignableFrom(node) && isApplicable();
+            return isApplicable();
         }
 
         @Nonnull
--- a/src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
+++ b/src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
@@ -50,7 +50,7 @@ THE SOFTWARE.
         <j:forEach var="g" items="${groups}">
           <j:forEach var="p" items="${g.permissions}">
             <j:if test="${descriptor.showPermission(p)}">
-              <td width="*">
+              <td width="*" data-implied-by-list="${descriptor.impliedByList(p)}" data-permission-id="${p.id}">
                 <f:checkbox name="[${p.id}]" tooltip="${p.group.title}/${p.name} for ${attrs.sid}" checked="${instance.hasExplicitPermission(attrs.sid,p)}"/>
               </td>
             </j:if>
@@ -126,7 +126,7 @@ THE SOFTWARE.
 
       <!-- template row to be used for adding a new row -->
       <j:set var="id" value="${h.generateId()}"/>
-      <tr id="${id}" style="display:none" class="permission-row">
+      <tr id="${id}" style="display:none" class="permission-row" name="__unused__">
         <local:row sid="__SID__" />
       </tr>
     </table>
@@ -139,13 +139,12 @@ THE SOFTWARE.
      <f:helpArea />
     </table>
     <script>
-      (function() {
-        <!-- place master outside the DOM tree so that it won't creep into the submitted form -->
-        var master = document.getElementById('${id}');
-        var table = master.parentNode;
-        table.removeChild(master);
-      
+      //<![CDATA[
+      Behaviour.specify("#${id}button", 'GlobalMatrixAuthorizationStrategy', 0, function(e) {
         makeButton($$('${id}button'), function (e) {
+          var master = document.getElementById('${id}');
+          var table = master.parentNode;
+
           <!-- when 'add' is clicked... -->
           var name = prompt("${%User or group name:}");
           if (name == null) {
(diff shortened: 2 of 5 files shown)


src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -139,13 +139,12 @@ THE SOFTWARE.
      <f:helpArea />
     </table>
     <script>
-      (function() {
-        <!-- place master outside the DOM tree so that it won't creep into the submitted form -->
```
daniel-beck: This is not actually necessary. An extra row in the submission doesn't really hurt, and since principals with no permissions assigned are actually ignored server-side, not doing this will not change the result.

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
```diff
@@ -187,7 +187,7 @@ public AuthorizationMatrixNodeProperty newInstance(StaplerRequest req, @Nonnull
 
         @Override
         public boolean isApplicable(Class<? extends Node> node) {
-            return Node.class.isAssignableFrom(node) && isApplicable();
+            return isApplicable();
```
daniel-beck: This only existed as a hack to prevent Kubernetes Plugin from showing the node property, as it passes a wrong class argument.

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -202,20 +202,64 @@ THE SOFTWARE.
...
+      //<![CDATA[
         <!-- validates the name -->
         Behaviour.specify("#${strategyid} TR.permission-row", 'GlobalMatrixAuthorizationStrategy', 0, function(e) {
-          FormChecker.delayedCheck("${descriptorPath}/checkName?value="+encodeURIComponent(e.getAttribute("name")),"GET",e.firstChild);
+          if (e.getAttribute('name') == '__unused__') {
+            return;
+          }
+          if (!e.hasAttribute('data-checked')) {
+            FormChecker.delayedCheck("${descriptorPath}/checkName?value="+encodeURIComponent(e.getAttribute("name")),"GET",e.firstChild);
+            e.setAttribute('data-checked', 'true');
```
daniel-beck: Since we add calls to `Behavior#applySubtree` on the `<table>`, not doing this attribute check would result in all rows being checked again through unrelated operations on the table. `#applySubtree` on the row doesn't work (and would still result in 1 unnecessary check each time), so do this instead.

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -202,20 +202,64 @@ THE SOFTWARE.
...
         </j:if>
+      //<![CDATA[
         <!-- validates the name -->
         Behaviour.specify("#${strategyid} TR.permission-row", 'GlobalMatrixAuthorizationStrategy', 0, function(e) {
-          FormChecker.delayedCheck("${descriptorPath}/checkName?value="+encodeURIComponent(e.getAttribute("name")),"GET",e.firstChild);
+          if (e.getAttribute('name') == '__unused__') {
+            return;
+          }
+          if (!e.hasAttribute('data-checked')) {
+            FormChecker.delayedCheck("${descriptorPath}/checkName?value="+encodeURIComponent(e.getAttribute("name")),"GET",e.firstChild);
```
Wadeck: ⚠️ if it was `delayedCheck('${descriptorPath` with single quote instead of double, it would have been XSS vulnerable.

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -202,20 +202,64 @@ THE SOFTWARE.
...
+                var list = td.getAttribute('data-implied-by-list');
+                if (list.includes(permissionId)) {
+                  inputs[i].disabled = checked;
+                }
+              }
+            }
+          }
+          return true;
+        };
+        e = null; <!-- avoid memory leak -->
```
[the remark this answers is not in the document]
daniel-beck: No idea. It's been in Git since the beginning, https://github.com/jenkinsci/jenkins/blob/1.312/core/src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly#L146, i.e. it's at least 10 years old. Perhaps related to browser bugs from a decade ago?

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
daniel-beck: Any way we can make this safer? `encodeURIComponent` this part, too?

Wadeck: The current version is safe thx to the Jelly escape. Nothing to change, that was just an "informational warning".

`EncodeURIComponent` will not solve the injection as the malicious code is applied during "jelly compilation" and the encode is called during JavaScript execution. If the malicious code put like `') + alert(1`, the encode will be just completely blind :)

PR #62, opened 2019-04-27
Properly work with optional dependency

daniel-beck:
An exception will occur during startup in 2.4 when Configuration-as-Code is not installed.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
+++ b/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
@@ -87,7 +87,8 @@ public class AuthorizationMatrixNodeProperty extends NodeProperty<Node> implemen
 
     private InheritanceStrategy inheritanceStrategy = new InheritGlobalStrategy();
 
-    private AuthorizationMatrixNodeProperty() {
+    @Restricted(NoExternalUse.class)
+    public AuthorizationMatrixNodeProperty() {
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
```diff
@@ -240,31 +241,5 @@ protected void onCreated(@Nonnull Node node) {
         }
     }
 
-    @Extension(optional = true)
-    @Restricted(NoExternalUse.class)
-    public static class Configurator extends BaseConfigurator<AuthorizationMatrixNodeProperty> {
```
jglick: Fix imports.

daniel-beck:
> Not sure what was going on here. Specific to Java 11? If so, maybe due to the new “nest” system.

No, 1.8.0_212-8u212-b01-1~deb9u1-b01 -- `docker run -p 8080:8080 jenkins/jenkins:2.164.2`, skip the wizard, and drop the 2.4 hpi into the upload form.

> harmful

Without alternative.

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationMatrixNodeProperty.java
daniel-beck: https://github.com/jenkinsci/matrix-auth-plugin/commit/a95020c41df2ab5e43422b4404af73b8673a1b4c

PR #63, opened 2019-05-02
Update AuthorizationContainerDescriptor.java

desillusion:
create false produces ugly NPEs in den GUI, that only resolve if the user logs into Jenkins.

Tested with Jenkins 2.164.2, Pluginversion 2.4.1

We use Atlassian Crowd as authentification provider!

npe


daniel-beck:
Thanks for bringing this to my attention! I'm going with the slightly different #64.

PR #65, opened 2019-07-26
Cache implying permissions for `hasPermission()` checks

AbhyudayaSharma:
Caches implying permissions for improving performance of `hasPermission()` checks.

This pull request follows from https://github.com/jenkinsci/role-strategy-plugin/pull/83 where we saw large improvements in performance of `hasPermission()` calls.

@jenkinsci/gsoc2019-role-strategy

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/matrixauth/ImplyingPermissionsCache.java
+++ b/src/main/java/org/jenkinsci/plugins/matrixauth/ImplyingPermissionsCache.java
@@ -1,0 +1,47 @@
+package org.jenkinsci.plugins.matrixauth;
+
+import hudson.security.Permission;
+
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Caches implying permissions for fast future access.
+ *
+ * The cache is built up when this class is loaded.
+ */
+class ImplyingPermissionsCache {
+    private static ConcurrentHashMap<Permission, Set<Permission>> implyingPermissionCache;
+
+    static {
+        implyingPermissionCache = new ConcurrentHashMap<>();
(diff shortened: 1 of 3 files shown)


oleg-nenashev:
Javadoc and Java 11 compatibility, hmm...

```
xit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/8/docs/api/ are in the unnamed module.

[2019-07-26T11:16:07.223Z] /home/jenkins/workspace/Plugins_matrix-auth-plugin_PR-65/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainer.java:76: warning: no @param for shortForm

[2019-07-26T11:16:07.223Z]     default void add(String shortForm) {

```

Newer plugin POM is needed IIUC

src/main/java/org/jenkinsci/plugins/matrixauth/ImplyingPermissionsCache.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.HashSet;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * Caches implying permissions for fast future access.
+ *
+ * The cache is built up when this class is loaded.
+ */
+class ImplyingPermissionsCache {
```
daniel-beck: Note to self: This isn't public, so probably no benefit to mark `@Restricted`.

daniel-beck:
@AbhyudayaSharma Could you add a test or similar demonstration of the benefits of this change? Is there's an actual improvement, it should be possible to show it. Otherwise I don't see the benefit of merging a PR that adds code and complexity. Thanks!

daniel-beck:
Closing after two months without a response, as general housekeeping.

I'd be happy to reconsider this submission once previous feedback is addressed.

PR #67, opened 2019-09-03
[JENKINS-59172] - Use GitHub as a source of the plugin's documentation on plugins.jenkins.io

oleg-nenashev:
There is nothing left on the Wiki Page except changelog, so we can safely replace it by GitHub page.

See https://groups.google.com/forum/#!topic/jenkinsci-dev/VSdfVMDIW-A


daniel-beck:
Please do not file origin branches.

PR #70, opened 2019-10-13
Add common badges to README

darxriggs:


daniel-beck:
Could you explain the benefit that these badges bring?

- CI status can be seen on the same screen as the readme.
- This is a formerly bundled, currently installed-by-default plugin, making the popularity stats entirely uninteresting.
- The latest version and popularity are shown on the plugin site, which is the usual landing page for the plugin anyway.

So the only benefit I see is that this adds a link from the GitHub repo to the plugin site, but it's pretty obscured (wouldn't expect the badges to be links).

darxriggs:
I see the README as made for developers and these badges and links provide a quick overview and fast access to relevant pages. Jenkins and other plugins also use similar badges. See some examples below.

- https://github.com/jenkinsci/jenkins/blob/master/README.md
- https://github.com/jenkinsci/git-client-plugin/blob/master/README.md
- https://github.com/jenkinsci/configuration-as-code-plugin/blob/master/README.md

I see your point about the number of installations vs. installed by default. But this can still be an indication for a developer that this plugin is quite important.

If you think the badges are not that useful, feel free to just close this pull request unmerged.

daniel-beck:
I think I will do that. Seems mostly redundant to me, plus

>README as made for developers

now that the GitHub README is also the user documentation shown on ci.jenkins.io, this seems like an audience mismatch. Most of that information is a click or so away anyway.

<details><summary>E.g. CI status</summary>
> Screenshot
</details>

Thanks anyway for considering this plugin for your contribution.

PR #74, opened 2020-01-16
Upgrade to use jcasc test harness

timja:
JCasc had to create a new dependency that wasn't a `tests` classifier because maven doesn't allow you to bring dependencies through classifiers.

Upgrading here so that PCT works again,

Tracking:
https://github.com/jenkinsci/bom/pull/164

Note: We'll need a release for PCT please

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,7 +77,7 @@
             <!-- requireUpperBoundDeps -->
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>structs</artifactId>
-            <version>1.13</version>
+            <version>1.19</version>
             <scope>test</scope>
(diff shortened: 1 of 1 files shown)


daniel-beck:
@timja Could you PTAL at JENKINS-60799 and see why the user claims he only gets 1.32? I'd rather not break even more Docker images with this change.

timja:
> JENKINS-60799

I don't understand that issue, install-plugins should install transitive dependencies automatically.

The issue that I could see happening would be users that have pinned configuration-as-code to an old version but updated a plugin that requires a newer version, and the error should be pretty obvious.

Also the issue has no error message to give context, so can't give more information than this really.

PR #78, opened 2020-02-25
JENKINS-61205 System read support

timja:
image

### DIFF
--- a/src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
+++ b/src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
@@ -22,7 +22,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 -->
 <?jelly escape-by-default='true'?>
-<j:jelly xmlns:j="jelly:core" xmlns:d="jelly:define" xmlns:f="/lib/form">
+<j:jelly xmlns:j="jelly:core" xmlns:d="jelly:define" xmlns:f="/lib/form" xmlns:l="/lib/layout">
   <f:block xmlns:local="local">
     <j:set var="groups" value="${descriptor.allGroups}"/>
     <d:taglib uri="local">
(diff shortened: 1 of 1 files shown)


src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -227,6 +233,12 @@ THE SOFTWARE.
         var tr = findAncestor(e,"TR");
         e.disabled = false;
         e.setAttribute('tooltip', findAncestor(e, "TD").getAttribute('data-tooltip-enabled'));
+
+        var submitButtonPresent = document.getElementsByClassName("submit-button").length > 0;
+        if (!submitButtonPresent) { // readonly mode if there's no submit button
```
daniel-beck:
```suggestion
        if (!submitButtonPresent) { // readonly mode if there's no submit button (Jenkins 2.222+)
```

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -135,9 +137,13 @@ THE SOFTWARE.
       </tr>
     </table>
     <table style="margin-top:0.5em; margin-left: 2em; width: 100%">
-     <tr><td colspan="3">
-      <input type="button" value="${%Add user or group…}" id="${id}button"/>
-     </td><td align="right">
+     <tr>
+       <td colspan="2">
```
daniel-beck: Reduced `colspan` why?

timja: This help icon is quite mis aligned, I didn’t fix it fully but it looks less bad with this change, I changed it because with the hiding of the add user / group button it looked worse where it was

timja: I've removed it from system read (and reverted the colspan), it was for the button that was being hidden.

(either colspan 2 or 3 both don't look good here, tbh UI could do with a re-work... )

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -227,6 +233,12 @@ THE SOFTWARE.
         var tr = findAncestor(e,"TR");
         e.disabled = false;
         e.setAttribute('tooltip', findAncestor(e, "TD").getAttribute('data-tooltip-enabled'));
+
+        var submitButtonPresent = document.getElementsByClassName("submit-button").length > 0;
+        if (!submitButtonPresent) { // readonly mode if there's no submit button (Jenkins 2.222+)
+            e.disabled = true;
+            return;
+        }
```
daniel-beck, line 241: Is this still what you would consider the preferred way to make this distinction, or would a check for `readOnlyMode` be better?

timja: Initially I had added a hidden input with the readOnly status in it. But you asked for it to be detected instead so I changed it to this which works well.

I wouldn’t want to use jelly vars in inline JavaScript really as it would make it more difficult to extract it

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -60,19 +60,21 @@ THE SOFTWARE.
...
-          <a href="#" class="selectall">
-            <img alt="${%Select all}" title="${%selectall(attrs.sid)}" src="${rootURL}/plugin/matrix-auth/images/16x16/select-all.png" height="16" width="16"/>
-          </a>
-          <a href="#" class="unselectall">
-            <img alt="${%Unselect all}" title="${%unselectall(attrs.sid)}" src="${rootURL}/plugin/matrix-auth/images/16x16/unselect-all.png" height="16" width="16"/>
-          </a>
-          <j:if test="${attrs.sid!='anonymous' and attrs.sid!='authenticated'}">
-            <a href="#" class="remove">
-              <img alt="${%Remove user/group}" title="${%remove(attrs.sid)}" src="${imagesURL}/16x16/stop.png" height="16" width="16"/>
+        <l:isAdmin>
```
daniel-beck, line 63: This form is reused for the project configuration 😬

PR #80, opened 2020-04-25
[SECURITY-410] Remove support for 'dangerous permissions'

daniel-beck:
It's been [three years since version 1.5 restricted the ability to configure "dangerous permissions"](https://github.com/jenkinsci/matrix-auth-plugin#version-15-apr-10-2017) for SECURITY-410.

[In Jenkins 2.222](https://www.jenkins.io/changelog/#v2.222), two months ago, those permissions were deprecated, and they also no longer work (at least for core features) as they have in the past when granted.

I think it makes sense to apply this general cleanup, simply dropping support via the 'escape hatch'.


daniel-beck:
> incompatibility metadata

Unsure about that — I only expect relatively few instances to have this escape hatch applied, so making 200k admins worry about upgrade safety doesn't seem like a great idea.

daniel-beck:
@oleg-nenashev To clarify, you'd just raise the core dependency to 2.222.x and done, since the permissions are no longer supported there anyway?

oleg-nenashev:
@daniel-beck 

> you'd just raise the core dependency to 2.222.x and done, since the permissions are no longer supported there anyway?

I would. https://www.jenkins.io/security/advisory/2020-03-25/ provides enough justification to use recent LTS versions, so bumping the version requirement seems perfectly fine to me

daniel-beck:
Strictly speaking 2.204.6 exists, but this looks like an easy pragmatic solution 👍

PR #82, opened 2020-05-07
[JENKINS-62202] Fix accidental read-only mode in jobs and folders

daniel-beck:
…gurations

----

[JENKINS-62202](https://issues.jenkins-ci.org/browse/JENKINS-62202)

Regression because of https://github.com/jenkinsci/matrix-auth-plugin/pull/78

### Testing notes

Enable permissions involved like this:

```
Item.EXTENDED_READ.enabled = true
Jenkins.SYSTEM_READ.enabled = true
Computer.EXTENDED_READ.enabled = true
```

`mvn hpi:run -Djenkins.version=2.235` helps some, but Computer/Configure can only be confirmed with https://github.com/jenkinsci/jenkins/pull/4531.

Manual testing setup:

1. `docker run --rm -ti -p 8080:8080 -e ID=4531 jenkins/core-pr-tester`
2. In Setup Wizard, install _only_ Folders Plugin
3. Create admin user
4. In Plugin Manager » Advanced, upload PR build
5. Create Manage Users, create user/user
6. In Script Console, run the script above

Now permissions can be assigned at different levels to ensure this works as intended.

### DIFF
--- a/src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
+++ b/src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
@@ -60,7 +60,7 @@ THE SOFTWARE.
             </j:if>
           </j:forEach>
         </j:forEach>
-        <l:isAdmin>
+        <l:hasPermission permission="${instance.getEditingPermission()}">
           <td class="stop" style="text-align:left;">
             <a href="#" class="selectall">
               <img alt="${%Select all}" title="${%selectall(attrs.sid)}" src="${rootURL}/plugin/matrix-auth/images/16x16/select-all.png" height="16" width="16"/>
@@ -74,7 +74,7 @@ THE SOFTWARE.
               </a>
             </j:if>
           </td>
-        </l:isAdmin>
+        </l:hasPermission>
       </d:tag>
     </d:taglib>
     <link rel="stylesheet" href="${rootURL}${app.VIEW_RESOURCE_PATH}/hudson/security/table.css" type="text/css" />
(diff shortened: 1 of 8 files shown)


src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -98,7 +98,9 @@ THE SOFTWARE.
             ${g.title}
           </td>
         </j:forEach>
-        <td rowspan="2" class="stop" />
+        <l:hasPermission permission="${instance.getEditingPermission()}">
+          <td rowspan="2" class="stop" />
+        </l:hasPermission>
```
daniel-beck, line 103: This fixes a visual artifact caused by the previous PR.

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -60,7 +60,7 @@ THE SOFTWARE.
             </j:if>
           </j:forEach>
         </j:forEach>
-        <l:isAdmin>
+        <l:hasPermission permission="${instance.getEditingPermission()}">
```
timja, line 63: I think you could use instance.editingPermission?

daniel-beck: 🤷 no difference there, and makes grepping slightly more straightforward.

timja:
Computer.EXTENDED_READ:

image

daniel-beck:
This doesn't work when viewing cloud agent templates for users with Computer/Configure and Overall/SystemRead but lacking Overall/Administer:

> Screenshot

Probably a small price to pay to get this fix, but still 😫

PR #83, opened 2020-05-10
[JENKINS-56109] Compatibility with table => div

timja:
https://github.com/jenkinsci/jenkins/pull/3895/files#r422546234

cc @daniel-beck 

Testing:
* test with: https://github.com/jenkinsci/jenkins/pull/3895 `docker run --rm -ti -p 8080:8080 -e ID=3895 jenkins/core-pr-tester`
* Set 'Project-based Matrix Authorization Strategy'
* Go to the `configure` page and make sure the 'Enable node-based security' optional property behaves properly

PR #86, opened 2020-09-15
[JENKINS-56109] Compatibility with upcoming div-based form layout

daniel-beck:
Supersedes #83, cleaner this way.

@timja PTAL


Wadeck (approved):
Nice approach! (not tested)

timja (approved):
nice

PR #92, opened 2020-11-05
Fix case sensitivity issue for LDAP logins

wirdman:
Fix case sensitivity issue between LDAP plugin and matrix-auth plugin. Basically logging in with UPERCASE or CamelCase is allowed in LDAP however the matrix-auth plugin does not apply the same permissions.


daniel-beck:
This shouldn't be a problem (and this certainly isn't the solution!) because this is just a performance shortcut. The final comparison is done a few lines further down when the case sensitivity strategy of the configured security realm is used.

Please report an issue in the Jenkins issue tracker with steps to reproduce the problem from scratch and provide your global `config.xml`.

PR #93, opened 2020-11-13
Making PermissionAdderTest run in PCT

jglick:
In https://github.com/jenkinsci/bom/pull/359 the test from #23 fails when `jth.jenkins-war.path` is set to the megawar. Seems that `SignupPage.submit` quietly returns even when the signup in fact fails. Trying the page interactively showed that it would refuse to proceed unless you entered an email address. Making this change makes the test pass when `jth.jenkins-war.path` is set. The test utility predates the new signup page GUI but it is not clear that is relevant; I have not figured out what about the megawar (bundled JavaScript?) would change behavior here.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
+++ b/src/test/java/org/jenkinsci/plugins/matrixauth/PermissionAdderTest.java
@@ -36,9 +36,11 @@ public class PermissionAdderTest {
                 signup.enterUsername("alice");
                 signup.enterPassword("alice");
                 signup.enterFullName("Alice User");
+                signup.enterEmail("alice@nowhere.net");
                 signup.submit(r.j);
-
-                Assert.assertTrue(r.j.jenkins.getACL().hasPermission(User.get("alice", false, Collections.emptyMap()).impersonate(), Jenkins.ADMINISTER));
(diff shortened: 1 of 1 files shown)


daniel-beck:
> I have not figured out what about the megawar (bundled JavaScript?)

Presence of mailer plugin?

jglick:
Oh, perhaps.

jglick:
Hmm, apparently needs to catch `ElementNotFoundException` in case `mailer` is absent. On mobile, can try that on Monday.

jglick:
https://github.com/jenkinsci/jenkins/blob/7aa52381f8d716f19ff71d7cf37970767ed5bbac/core/src/main/resources/hudson/security/HudsonPrivateSecurityRealm/_entryForm.jelly#L75 FTR

daniel-beck:
Now it failed in https://github.com/jenkinsci/matrix-auth-plugin/blob/df6f6491f1d9ddab40fd4476852d8c9940c5762f/src/test/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixPropertyTest.java#L111, rerunning.

jglick:
Hmm, passed locally, not sure what that is about.

PR #99, opened 2021-01-28
Add HTTP verb annotations to all web methods

daniel-beck:
Essentially an experiment, but should work just fine.

Getting explicit verb annotations even without https://github.com/stapler/stapler/pull/205


src/main/java/com/cloudbees/hudson/plugins/folder/properties/AuthorizationMatrixProperty.java
```diff
@@ -171,6 +172,7 @@ public boolean isApplicable(Class<? extends AbstractFolder> folder) {
         }
 
         @GET
+        @POST
```
jglick, line 175: Delete `@GET` and set `checkMethod="post"`.

jglick (changes requested):

PR #105, opened 2021-11-18
[JENKINS-67210] Fix linking to the global security realm

NotMyFault:
The placeholder at the "Inherit permissions from parent ACL" tab is not resolved, so it would forward you to `localhost:8080/job/jobName/%7B0%7D/configureSecurity` instead of `localhost:8080/configureSecurity`.
Considering it's omitted in the other tabs, stripping it resolves named issue while maintaining a working way for the other tabs.

- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Ensure you have provided tests - that demonstrates feature works or fixes the issue


daniel-beck:
This won't work for people who use a non-empty context path. Callers need to be fixed.

daniel-beck:
https://issues.jenkins.io/browse/JENKINS-67210 identifies the problem; it's not with callers, but rather an improper escape character in the localized string. #106 fixes this, while retaining support for non-empty context paths.

Thanks for making me aware of the bug, and this fix proposal!

PR #107, opened 2021-11-24
[JENKINS-66964] Fix button tooltips

daniel-beck:
See [JENKINS-66964](https://issues.jenkins.io/browse/JENKINS-66964)

CC @Kevin-CB


Wadeck (approved):
Alternative working solution:

inside `<d:tag name="row">`
Put somewhere:
`<j:set var="sid" value="${attrs}" />`

The problem is that the `attrs` is "re"defined by Jelly when executing a tag and thus removing the one from the parent tag.

PR #110, opened 2021-12-08
[JENKINS-67326] Explicitly assign permissions by type (user/group)

daniel-beck:
See [JENKINS-67326](https://issues.jenkins.io/browse/JENKINS-67326)

I'm not super happy with the current version of the Job DSL and JCasC API, those can probably be done much nicer later. For now, this is functional.


timja (approved):
Looks good, I've done quite a bit of testing on this.

Last change I just did was adapt my `getPrincipal()` to match what was being stored, previously it worked because the ACL checked all authorities but now it's checking based on a user / group.

My PR is https://github.com/jenkinsci/azure-ad-plugin/pull/177 if you are interested (very small PR compared to this)

ccayg-sainsburys:
Am I misunderstanding something here?
I updated to v3 after reading the changelog.
I then saw 'No type prefix: ' for all users in /role-strategy/assign-roles
After looking at this PR plus a little trial and error I added for example 'USER:admin'
That seemed to link up as the display name was then shown with a 'person icon' and the warning was cleared.
BUT that user now is not receiving the permissions of the role assigned to it.

Does this only work for specific types of users?
Am I doing something wrong is the responsibility for this in a different plugin?

daniel-beck:
@ccayg-sainsburys 

>  I then saw 'No type prefix: ' for all users in /role-strategy/assign-roles

`role-strategy` has not yet been adapted. The release notes state:

> ## Important upgrade notes
> 1. …
> 2. Plugin APIs have changed significantly.
> While some compatibility is retained, other plugins that depend on this plugin will likely need to be adapted to these changes or may behave in unexpected ways.
> **If you use any plugins with a dependency on this plugin, make sure they're compatible with this release before upgrading.**

ccayg-sainsburys:
> > I then saw 'No type prefix: ' for all users in /role-strategy/assign-roles
> 
> `role-strategy` has not yet been adapted and I believe it needs to me. The release notes state this requirement:

Yes, release notes seen - I hadn't understood the different plugin responsibilities though. Thanks for the reply.

Conrad-T-Pino:
@ccayg-sainsburys!
`USER:admin` was puzzle piece not found and experiment not done.
Thank you, thank you, thank you ...

Conrad-T-Pino:
@daniel-beck
Please comment on [JENKINS-67426 PermissionEntry Persistence Support Method](https://issues.jenkins.io/browse/JENKINS-67426) when convenient.
I want to use `PermissionEntry` in Role-based Authorization [`RoleMap`](https://github.com/Conrad-T-Pino/role-strategy-plugin/blob/master/src/main/java/com/michelin/cio/hudson/plugins/rolestrategy/RoleMap.java) to resolve UI warning and persistence.

z0rc:
What the new format for jobDSL's `authorization {}` block (https://jenkinsci.github.io/job-dsl-plugin/#path/javaposse.jobdsl.dsl.jobs.IvyJob.authorization)? Documentation for `permission(String permission)`, `permission(String permission, String user)`, `permissionAll(String user)` and `permissions(String user, Iterable<String> permissionsList)` method needs an update. 

I have multiple blocks of that look like this:
```
job('example-4') {
    authorization {
        permissions('Engineering', [
            'hudson.model.Item.Create',
            'hudson.model.Item.Discover'
        ])
    }
}
```

Tried to change it to `permissions('GROUP:Engineering', ...`, but this created rule with literal `GROUP:Engineering` entity.

z0rc:
@daniel-beck thanks, I updated my code to use `properties { authorizationMatrix { permissions () } }` model, as seen in provided example.

PR #113, opened 2022-01-03
Address sunset icon removal from core

NotMyFault:
A possible solution for https://github.com/jenkinsci/jenkins/pull/5778#issuecomment-1004180179.
Versions lower than 2.308, the first GA version SVG icons have been introduced in, use the PNG icon path, versions newer will pick the SVG icon path.

Just a quick proposal that came into my mind.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptor.java
+++ b/src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptor.java
@@ -161,10 +161,10 @@ public interface AuthorizationContainerDescriptor<T extends AuthorizationContain
         if (!subject.hasPermission(permission)) {
             // Lacking permissions, so respond based on input only
             if (type == AuthorizationType.USER) {
-                return FormValidation.okWithMarkup(formatUserGroupValidationResponse("person.png", escapedSid, "User may or may not exist", false));
+                return FormValidation.okWithMarkup(formatUserGroupValidationResponse("person", escapedSid, "User may or may not exist", false));
             }
             if (type == AuthorizationType.GROUP) {
-                return FormValidation.okWithMarkup(formatUserGroupValidationResponse("user.png", escapedSid, "Group may or may not exist", false));
+                return FormValidation.okWithMarkup(formatUserGroupValidationResponse("user", escapedSid, "Group may or may not exist", false));
             }
             return FormValidation.warningWithMarkup(formatUserGroupValidationResponse(null, escapedSid, "Permissions would be granted to a user or group of this name", false));
         }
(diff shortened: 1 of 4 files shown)


daniel-beck:
@NotMyFault Thanks for the further enhancement!

I'm curious, where did you get the "disabled user" SVG from? It's not https://raw.githubusercontent.com/RRZE-PP/rrze-icon-set/master/tango/scalable/categories/user-disabled.svg as I would have expected given the 1:1 updated attribution.

Could you update the check/uncheck all icons to have multiple checkboxes? I think the PNG icons better represented the action to select/deselect all as there were two (stacked) checkboxes shown. Of course if you don't feel like it, the latest changes can always be backed out and done some other time :)

NotMyFault:
> I'm curious, where did you get the "disabled user" SVG from? It's not https://raw.githubusercontent.com/RRZE-PP/rrze-icon-set/master/tango/scalable/categories/user-disabled.svg as I would have expected given the 1:1 updated attribution.

I suspect some chrome shenanigans here, the icon looks perfectly fine in Firefox, as it looks on Jenkins


> Could you update the check/uncheck all icons to have multiple checkboxes?
Yeah, I'm looking forward to do so in a bit.

NotMyFault:
Aight, the checkboxes are now doubled again.

daniel-beck:
> I suspect some chrome shenanigans here, the icon looks perfectly fine in Firefox, as it looks on Jenkins

To clarify, the file in this PR has some Inkscape metadata, while the one I link to does not. Even if they look alike, they are different in some (undocumented) manner? Did you post-process them?

> Aight, the checkboxes are now doubled again.

Thanks! I assume these look reasonable on Dark Theme as well with the transparency?

NotMyFault:
> Thanks! I assume these look reasonable on Dark Theme as well with the transparency?

Yeah, they are transparent.

> To clarify, the file in this PR has some Inkscape metadata, while the one I link to does not. Even if they look alike, they are different in some (undocumented) manner? Did you post-process them?

Yeah, I adjusted its viewpoints a bit looking better in the table view and reduced the overall line count by ~100.
If that is undesired, I can revert it for sure.

daniel-beck:
> Yeah, I adjusted its viewpoints a bit looking better in the table view and reduced the overall line count by ~100.

Great, thanks.

> If that is undesired, I can revert it for sure.

Nope, all good. I just noticed the difference in review and wondered about it. Seems compatible with the license.

NotMyFault:
@daniel-beck mind taking a look? 👀

daniel-beck:
## Old Jenkins
Screenshot 2022-01-20 at 15 10 00

## Current LTS
Screenshot 2022-01-20 at 15 06 57

## Current Weekly
Skipping, Artifactory is acting up 😭 

Looks good 👍 Is it safe to assume I don't need to try it with https://github.com/jenkinsci/jenkins/pull/5778? The only non SVG stuff I see in 2.319.2 is provided by core.

NotMyFault:
> Looks good 👍 Is it safe to assume I don't need to try it with jenkinsci/jenkins#5778? The only non SVG stuff I see in 2.319.2 is provided by core.

Yeah. The non svgs in the table are addressed by Jan's PR as well, but that is no issue for now.

daniel-beck:
Will release once I have confirmation that Artifactory is stable again.

NotMyFault:
According to timja, Artifactory seems sorted, https://github.com/jenkins-infra/helpdesk/issues/2733#issuecomment-1017614448, fingers crossed it stays like that 🤞🏻

daniel-beck:
https://github.com/jenkinsci/matrix-auth-plugin/releases/tag/matrix-auth-3.0.1

PR #116, opened 2022-04-13
[JENKINS-67902] Fix tooltips not showing on checkboxes 

janfaracik:
Relating to https://issues.jenkins.io/browse/JENKINS-67902.

The problem was was that in newer versions of Jenkins the checkbox `input` is invisible, meaning that there is no hitbox for the user to show the tooltip. The solution is to instead apply the tooltip to the `label`, which due to CSS shenanigans actually renders the checkbox and its hitbox.  

- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

### DIFF
--- a/src/main/resources/hudson/security/table.js
+++ b/src/main/resources/hudson/security/table.js
@@ -180,7 +180,7 @@ Behaviour.specify(".global-matrix-authorization-strategy-table td input", 'Globa
   var impliedByList = impliedByString.split(" ");
   var tr = findAncestor(e,"TR");
   e.disabled = false;
-  e.setAttribute('tooltip', YAHOO.lang.escapeHTML(findAncestor(e, "TD").getAttribute('data-tooltip-enabled')));
+  e.nextSibling.setAttribute('tooltip', YAHOO.lang.escapeHTML(findAncestor(e, "TD").getAttribute('data-tooltip-enabled')));
 
   for (var i = 0; i < impliedByList.length; i++) {
     var permissionId = impliedByList[i];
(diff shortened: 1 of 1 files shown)


daniel-beck:
Will this still work on the older core baselines this plugin claims to be compatible with, or do I need to depend on a recent weekly?

janfaracik:
> Will this still work on the older core baselines this plugin claims to be compatible with, or do I need to depend on a recent weekly to not break tooltips for users on slightly older core releases?

Ah it won't work on the latest LTS, earliest version it would work on is 2.335 due to the changed checkboxes.

daniel-beck:
Ideally there would be conditional behavior (e.g. checking visibility? Unsure) that makes it work on all versions of Jenkins. I'd rather not have to go with a 2.335 weekly baseline for the plugin because of this.

basil:
Daniel, we have a lot of regressions to fix, so if you are willing to make an exception and accommodate a higher baseline in order to expedite the delivery of this fix, that would definitely speed up the process. I don't deny that it would be better to support multiple versions of Jenkins, but we have a long list of regressions to work on, and spending more time on this one is just going to delay progress on the others.

daniel-beck:
This fix (as a result of the core change I guess) also addresses [JENKINS-63391](https://issues.jenkins.io/browse/JENKINS-63391).

janfaracik:
> @janfaracik Do you see any problems with the hack I added to this PR?

Looks good to me! :) Nice fix

src/main/resources/hudson/security/table.js
```diff
@@ -180,15 +180,19 @@ Behaviour.specify(".global-matrix-authorization-strategy-table td input", 'Globa
   var impliedByList = impliedByString.split(" ");
   var tr = findAncestor(e,"TR");
   e.disabled = false;
-  e.setAttribute('tooltip', YAHOO.lang.escapeHTML(findAncestor(e, "TD").getAttribute('data-tooltip-enabled')));
+  let tooltip = YAHOO.lang.escapeHTML(findAncestor(e, "TD").getAttribute('data-tooltip-enabled'))
```
basil, line 183:
```suggestion
  let tooltip = YAHOO.lang.escapeHTML(findAncestor(e, "TD").getAttribute('data-tooltip-enabled'));
```

basil (approved):

PR #118, opened 2022-04-21
Remove 'disabled user' icon on lookup error

daniel-beck:
#110 switched from `FormValidation.respond(ERROR, …)` to `FormValidation.errorWithMarkup(…)` and as a side effect, we're now using the core-provided format of red, bold text and error icon. This makes the "not found" user icon redundant.

As removal is easier than switching the presentation based on core version for compatibility, do the former. Thanks to @NotMyFault for the idea.

## Before

> Screenshot

## After

> Screenshot


NotMyFault (approved):
Nice one 👏🏻

PR #120, opened 2022-05-02
Fix tooltips with Tippy.js

janfaracik:
This PR fixes tooltips for Jenkins [#6408](https://github.com/jenkinsci/jenkins/pull/6408) which replaces YUI tooltips with Tippy.js ones.

Essentially the fix is to reregister the tooltips once a new item has been added otherwise Tippy won't know about the new element.

Thanks @timja :) 

---

- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue


timja (approved):
I tested on 2.277.1 and with https://github.com/jenkinsci/jenkins/pull/6408

Works fine and looks much better with the new core PR

daniel-beck:
> Screenshot 2022-05-02 at 21 43 36

Is this a side effect of the new non-HTML by default behavior?

timja:
> Would window.registerTooltips be how I check for the presence of Tippy?

Yes

daniel-beck (changes requested):
Checking and unchecking a checkbox takes seconds to apply. There needs to be a better solution than this.

daniel-beck (changes requested):
Checking and unchecking a checkbox takes seconds to apply. There needs to be a better solution than this.

(I forgot that I restarted with 2.345 rather than the incremental, so everything was fast 😑)

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationContainerDescriptor.java
```diff
@@ -132,7 +132,7 @@ default PermissionEntry entryFor(String type, String sid) {
     @Restricted(DoNotUse.class) // Jelly only; cf. UpdateCenter#getCategoryDisplayName in core
     default String getTypeLabel(String type) throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
         if (type == null) {
-            return null;
+            return "__TYPE__";
```
timja, line 135: new values were showing null before after clicking add, this isn't fully fixed as it's not localised yet, but I thought I'd push it for now

src/main/resources/hudson/security/table.js
```diff
@@ -238,3 +246,9 @@ Behaviour.specify(".global-matrix-authorization-strategy-table TR.permission-row
     e.setAttribute('data-checked', 'true');
   }
 });
+
+document.addEventListener('DOMContentLoaded', function () {
+  setTimeout(function () {
+    window.registerTooltips();
+  }, 500);
```
timja, line 253: quite hacky but haven't managed to find a way yet to call this after all inputs have had their javascript applied.

the previously implementation was being called ~220 times with ~5 or so entries. which is why it was so slow

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -178,27 +178,29 @@ THE SOFTWARE.
...
-            data-message-error="${%userError}"
-      />
-      <input type="button" class="matrix-auth-add-button" value="${%Add group…}" id="${id}GroupButton"
-            data-table-id="${id}"
-            data-type="GROUP"
-            data-message-prompt="${%groupPrompt}"
-            data-message-empty="${%groupEmpty}"
-            data-message-error="${%groupError}"
-      />
+      <div class="jenkins-!-margin-top-2">
```
timja, line 181: minor spacing fix, can extract if needed

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -178,27 +178,29 @@ THE SOFTWARE.
...
+        />
+        <input type="button" class="matrix-auth-add-button" value="${%Add group…}" id="${id}GroupButton"
+              data-table-id="${id}"
+              data-type="GROUP"
+              data-message-prompt="${%groupPrompt}"
+              data-message-empty="${%groupEmpty}"
+              data-message-error="${%groupError}"
+        />
+        <f:helpLink featureName="${%Permissions matrix}" url="${descriptor.find('hudson.security.GlobalMatrixAuthorizationStrategy$DescriptorImpl').getHelpFile('user-group')}"/>
+        <f:helpArea />
```
timja, line 197: fixes display of help link if ambiguous warning is on the page

daniel-beck:
Nice, the current state no longer opens the browser built-in tooltip after clicking a checkbox.

Is it necessary to have the tooltips interactable? With a grid like matrix-auth's, it's annoying to move the pointer up and rather than disappear and allowing me to interact with the row of checkboxes, the tooltip is now itself the hovered element (and I can select text etc. from it).

timja:
> Is it necessary to have the tooltips interactable?

hmm looking at the demo on https://atomiks.github.io/tippyjs/ it doesn't do that, looking

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -146,7 +146,11 @@ THE SOFTWARE.
         <j:forEach var="g" items="${groups}">
           <j:forEach var="p" items="${g.permissions}">
             <j:if test="${descriptor.showPermission(p)}">
-              <th class="pane" tooltip="${descriptor.getDescription(p)}">
+              <th class="pane" html-tooltip="${descriptor.getDescription(p)}" tooltip="${descriptor.getDescription(p)}">
+                <!--
+                  Tooltip needs to be HTML for GlobalMatrixAuthorizationStrategy.PermissionNotImpliedBy formatting.
+                  So set 'html-tooltip' for Tippy and 'tooltip' for YUI.
+                 -->
```
daniel-beck, line 153: Can I rely on `html-tooltip` to have precedence over `tooltip` for this situation?

timja: no they'll both get set from what I can tell, probably want to correct it in JavaScript.

timja:
> I assume it's the `interactive: true,` in the `tooltips.js` file

tested and yes it's that, it behaves how you expect without it, asked on https://github.com/jenkinsci/jenkins/pull/6408/files#r863493271

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
daniel-beck: Any suggestions how I can make this compatible both with current Jenkins and future Jenkins?

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
daniel-beck: Is this CSS class one that I can rely on existing in the future?

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
timja: > correct it in JavaScript

^.^

https://github.com/jenkinsci/matrix-auth-plugin/pull/120/commits/3a215df7cecc7a018539fd20b6fdc42fe1b69c47

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
timja: Yes it's a stable API, documented at:
https://weekly.ci.jenkins.io/design-library/Spacing/

timja:
@daniel-beck any chance of another review please, thanks!

janfaracik:
Heya @daniel-beck, any update on this? Cheers :)

NotMyFault (approved):

daniel-beck:
> > Would window.registerTooltips be how I check for the presence of Tippy?
> 
> Yes

That didn't survive a code reorg in the core PR: https://github.com/jenkinsci/matrix-auth-plugin/pull/130#issuecomment-1342439887

src/main/resources/hudson/security/table.js
```diff
@@ -215,11 +231,16 @@ Behaviour.specify(".global-matrix-authorization-strategy-table td input", 'Globa
...
-        e.nextSibling.setAttribute('tooltip', tooltip); // 2.335+
+        e.setAttribute(tooltipAttributeName, tooltip); // before 2.335 -- TODO remove once baseline is new enough
+        e.nextSibling.setAttribute(tooltipAttributeName, tooltip); // 2.335+
       }
     }
   }
+
+  if (window.registerTooltips) {
+    window.registerTooltips(e.nextSibling.parentElement);
+  }
```
daniel-beck, line 242: What was the purpose of this? It doesn't work since https://github.com/jenkinsci/jenkins/pull/7464 renamed the method, and the other occurrence of `registerTooltips` was fixed up in https://github.com/jenkinsci/matrix-auth-plugin/pull/130, but this code still exists.

timja: I think it was fixed in core itself with behaviour and then applying subtree here.

the purpose was so that when checkboxes were dynamically added they would have tippy run on this but it was fixed in another way...

PR #121, opened 2022-05-03
Fix type in tooltip of newly added rows

daniel-beck:
Noticed during testing of https://github.com/jenkinsci/matrix-auth-plugin/pull/120, tooltips of newly added rows have a "null" type.

Thanks to @timja for getting this fix started.


timja (approved):
nice

PR #123, opened 2022-06-24
Update parent pom to 4.41

timja:
see https://github.com/jenkinsci/bom/pull/1133#issuecomment-1165846002

(and can this be released please?)

PR #124, opened 2022-06-24
Enable dependabot

timja:
Would you consider enabling dependabot here?


daniel-beck:
What would that accomplish? https://github.com/jenkinsci/solarized-theme-plugin/pulls is beyond silly, so I'd rather not duplicate that experience here.

timja:
A more up to date parent pom so bom works with new core changes.

timja:
The PR I'm actually interested in is https://github.com/jenkinsci/matrix-auth-plugin/pull/123

If you aren't interested then feel free to close

PR #130, opened 2022-09-30
[JENKINS-70231] Fix tooltips on 2.379+

janfaracik:
Adapts to https://github.com/jenkinsci/jenkins/commit/61f6553844b671a3a0dff5454464f90f3ac466c3

### DIFF
--- a/src/main/resources/hudson/security/table.js
+++ b/src/main/resources/hudson/security/table.js
@@ -201,7 +201,7 @@ Behaviour.specify(".global-matrix-authorization-strategy-table TD.stop A.migrate
 
 function getTooltipAttributeName() {
   var tippySupported = window.registerTooltips !== undefined;
-  return tippySupported ? 'html-tooltip' : 'tooltip';
+  return tippySupported ? 'data-html-tooltip' : 'tooltip';
 }
 
 /*
--- a/src/main/resources/hudson/security/table.js
+++ b/src/main/resources/hudson/security/table.js
@@ -212,6 +212,7 @@ function getTooltipAttributeName() {
   if (coreVersion === null) {
     return 'tooltip'
   }
+  // TODO remove after minimum version is 2.379 or higher
   let tippySupported = coreVersion >= '2.379';
   return tippySupported ? 'data-html-tooltip' : 'tooltip';
 }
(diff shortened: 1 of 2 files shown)


janfaracik:
Ping @daniel-beck

timja:
@daniel-beck do you have an ETA for this? it's currently holding up the tippy core PR

NotMyFault (approved):

daniel-beck:
Cannot test this PR on 2.380, probably due to [JENKINS-70178](https://issues.jenkins.io/browse/JENKINS-70178).

daniel-beck:
This PR appears to break the "Add User" and "Add Group" buttons on 2.380+. If you click them, nothing happens after entering the name. No errors are logged to the console.

Additionally, the tooltips show HTML entities, indicating this doesn't work as expected:

* 2.379 + matrix-auth 3.1.5: `Overall/Administer`
* 2.380 + matrix-auth 3.1.5: `Overall&#x2F;Administer` (weekly.ci)
* 2.379 + this PR: `Overall/Administer`
* 2.380 + this PR: `Overall&#x2F;Administer` (unexpected)

daniel-beck:
Found the problem of the PR breaking on 2.380+ and fixed it, but still unclear why the HTML entities are visible.

daniel-beck:
> `body[data-version]` maybe? Now that it's released…

This seems to work in a quick test, pushed a commit with the proposed fix.

src/main/resources/hudson/security/table.js
```diff
@@ -201,9 +204,16 @@ Behaviour.specify(".global-matrix-authorization-strategy-table TD.stop A.migrate
...
+ * Determine which attribute to set tooltips in. Changed in Jenkins 2.379 with Tippy and data-html-tooltip support.
+ */
 function getTooltipAttributeName() {
-  var tippySupported = window.registerTooltips !== undefined;
-  return tippySupported ? 'html-tooltip' : 'tooltip';
+  let coreVersion = document.body.getAttribute('data-version');
+  if (coreVersion === null) {
+    return 'tooltip'
+  }
+  let tippySupported = coreVersion >= '2.379';
```
timja, line 215: I would think about adding a TODO, this check could possibly break with interesting version numbers
```suggestion
  // TODO remove after minimum version is 2.379 or higher
  let tippySupported = coreVersion >= '2.379';
```

daniel-beck: If we get to 2.1000, yes 😁

timja: heh yeah that was the only one I found it broke on I did try quite a few in the browser

daniel-beck: It's an ascii-left-to-right string comparison. If we limit it to version numbers, the sort order is like in https://stats.jenkins.io/plugin-installation-trend/installations.json

PR #132, opened 2022-11-12
Compatibility with jenkins-buttons (2.378+)

NotMyFault:
Adopts https://github.com/jenkinsci/jenkins/pull/7364

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/matrixauth/AmbiguityMonitor/message.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/matrixauth/AmbiguityMonitor/message.jelly
@@ -26,7 +26,7 @@ THE SOFTWARE.
 <j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
     <div class="alert alert-warning">
         <form method="post" action="${rootURL}/${it.url}/disable">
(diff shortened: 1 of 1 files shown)


daniel-beck:
These classes aren't new though, right? So anyone on older Jenkins releases, including current and upcoming LTS, will have inconsistent admin monitor presentations?

NotMyFault:
Jenkins buttons exist for quite a while yet, although there's no coherent logic for a dismiss component yet, just a draft PR for core. Plugin authors used to handle the things on their own previously.

daniel-beck:
Right, but before your PR, Jenkins core admin monitors did not use the destructive buttons. So it's not "forward compatibility" as in the HTML tooltip PR compatibility PRs, but rather as in Jenkins core 2.378+ behavior, regardless of the version of Jenkins this plugin is installed on. Or what am I missing?

NotMyFault:
>  that result in the expected behavior on any core version, but rather as in Jenkins core 2.378+ behavior,

You are correct. Jenkins versions starting with the introduction of this jenkins-button component ranging to 2.378 would display things differently, if you merge this PR.

NotMyFault:
Closing, because the core PR is about to get reverted.

PR #138, opened 2023-04-18
Bump parent pom

timja:
Required for https://github.com/jenkinsci/jenkins/pull/7781 / https://github.com/jenkinsci/bom/pull/1968

Could I get a release please?

The parent pom bump includes https://github.com/jenkinsci/jenkins-test-harness/pull/567 which is released in https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.58

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>4.52</version>
+        <version>4.60</version>
     </parent>
     <artifactId>matrix-auth</artifactId>
     <version>${revision}${changelist}</version>


daniel-beck:
I considered that for https://github.com/jenkinsci/matrix-auth-plugin/pull/135 but the generated `pom.xml` looked weird to me. Do you understand why the dependency on `jenkins-core` is gone?

timja:
> I considered that as an addition to #135, but the generated `pom.xml` looked weird to me. Do you understand why the dependency on `jenkins-core` is gone (new in 4.57)?

Unsure is it causing any issues?

A bunch of maven plugins were updated in 4.57: https://github.com/jenkinsci/plugin-pom/releases/tag/plugin-4.57

daniel-beck:
> Unsure is it causing any issues?

I don't know, I'm not using `matrix-auth` as a dependency, where the pom would matter.

timja:
> > Unsure is it causing any issues?
> 
> I don't know, I'm not using `matrix-auth` as a dependency, where the pom would matter.

I tested it here with no issues: https://github.com/jenkinsci/azure-ad-plugin/pull/401

basil:
> I considered that as an addition to #135, but the generated `pom.xml` looked weird to me. Do you understand why the dependency on `jenkins-core` is gone (new in 4.57)?

I assume you mean this:

```xml
@@ -43,34 +42,10 @@
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.main</groupId>
-      <artifactId>jenkins-core</artifactId>
-      <version>2.361.4</version>
-      <scope>provided</scope>
-      <exclusions>
-        <exclusion>
-          <artifactId>servlet-api</artifactId>
-          <groupId>javax.servlet</groupId>
-        </exclusion>
-      </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>javax.servlet</groupId>
-      <artifactId>javax.servlet-api</artifactId>
-      <version>3.1.0</version>
-      <scope>provided</scope>
-    </dependency>
-    <dependency>
-      <groupId>commons-logging</groupId>
-      <artifactId>commons-logging</artifactId>
-      <version>1.2</version>
-      <scope>provided</scope>
-    </dependency>
   </dependencies>
   <repositories>
     <repository>
```

I assume one of the changes in https://github.com/mojohaus/flatten-maven-plugin/releases/tag/1.4.0 caused this, not sure which one. But that was a fairly significant release of Flatten Maven Plugin with a lot of cleanup so I am not surprised. In any case I think the change is desirable and an improvement rather than a bug. The "BOM" profile we use in Flatten Maven Plugin is designed to strip any unnecessary elements, and these are certainly unnecessary since they are not used at runtime. Besides which, as @timja noted, if plugin POM 4.57 really was causing problems we would likely have noticed it by now.

PR #141, opened 2023-05-30
[JENKINS-71368] Obtain groups from all matrix properties using SYSTEM identity

TWestling:
Using SYSTEM2 ACLContext to find out which group permissions has been set

The getGroups method of ProjectMatrixAuthorizationStrategy uses the user's ACLContext to iterate through all projects/folders/nodes and find which group permissions are set on each.

In conjunction with the active-directory plugin, with the option "remove irrelevant groups" this causes a problem, since we save the relevant groups upon logging in, when the user is still anonymous. This is done for performance reasons, since going through each AD group everytime a user tries to access a resource is not doable for systems with a lot of AD groups.

This means that for everything on the master where Anonymous doesn't have read access, the groups defined are deemed irrelevant, since we can't access them at the time of calculation. This in turn means that you can't set up a system where access is given through AD groups and anonymous can't already read everything.

Thus, changed to using the SYSTEM2 ACLContext to find out which groups are used.

Jira issue: [JENKINS-71368](https://issues.jenkins.io/browse/JENKINS-71368)

### Testing done

Tested in conjunction with the active-directory plugin with remove irrelevant groups checked.
Noticed that the groups can now be found, and the correct permissions can be set and used.
### Submitter checklist

- [*] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [*] Ensure that the pull request title represents the desired changelog entry
- [*] Please describe what you did
- [*] Link to relevant issues in GitHub or Jira
- [*] Link to relevant pull requests, esp. upstream and downstream changes
- [*] Ensure you have provided tests - that demonstrates feature works or fixes the issue


daniel-beck:
Thanks, this looks useful. Is there a JENKINS issue for this? Did you consider setting the authentication in the caller, and why did you decide it's better off here?

TWestling:
I don't know of any JENKINS issue for this. Should I create one?
I didn't actually consider setting the authentication in the active-directory and now that you mention it, maybe it is better off there, it's where the issue happens in this perfect storm. My thinking was that there was no reason that you shouldn't get all of the groups by calling getGroups, but I guess each caller could set it themselves if needed.

daniel-beck:
> I don't know of any JENKINS issue for this. Should I create one?

Not strictly necessary, but your description looks like a bug you encountered.

> My thinking was that there was no reason that you shouldn't get all of the groups by calling getGroups, but I guess each caller could set it themselves if needed.

Right; at the moment I'm leaning the same way. I don't expect this to have a security impact either.

https://javadoc.jenkins.io/hudson/security/AuthorizationStrategy.html#getGroups() seems unnecessarily narrowly described.

TWestling:
Correct, this was a bug that was found inside my company, but no jira was created on the Jenkins project since I imagined having to solve it myself anyway. Created https://issues.jenkins.io/browse/JENKINS-71368 now.

daniel-beck:
@TWestling I expect to create a new release with this change later this month.

TWestling:
Thanks!

daniel-beck (approved):

daniel-beck:
@TWestling WDYT, does the test look like it would prevent regressions?

daniel-beck:
Going ahead with merging, as the only open question is around non-production code.

PR #144, opened 2023-07-03
Implement nicer property DSL for Job DSL and Pipeline

daniel-beck:
Companion PR to #145 (which is for JCasC).

----
Job DSL:

```
folder('generated') {
  properties {
    authorizationMatrix {
      inheritanceStrategy {
          nonInheriting()
      }
      entries {
        group {
          name('authenticated')
          permissions([
            'Job/Build',
            'Job/Configure',
            'Job/Read'
          ])}
        user {
          name('admin')
          permissions(['Job/Delete'])
        }
      }
    }
  }
}
```

The corresponding Pipeline snippet looks like this, according to Snippet generator (formatted for readability):

```
properties([
  authorizationMatrix([
    user(name: 'admin', permissions: ['Job/Delete']),
    group(name: 'authenticated', permissions: [
      'Job/Build',
      'Job/Configure',
      'Job/Read'
    ])
  ])
])
```

----

I'm kinda worried about the symbols here though because of `configuration-as-code` incorrectly implementing the symbol scopes (hence `mavenGlobalConfig`). Is that still the case @timja ?


timja:
> I'm kinda worried about the symbols here though because of configuration-as-code incorrectly implementing the symbol scopes (hence mavenGlobalConfig). Is that still the case @timja ?

JCasC doesn't work with colliding symbols in the same category, e.g. 2 `maven` under globalconfiguration from what I remember.

I think what you've got is fine but we can run through bom tests?

daniel-beck:
> 2 `maven` under globalconfiguration from what I remember.

Per https://github.com/jenkinsci/jenkins/pull/5008, the symbols of the global configuration and the build step collided. That seemed wrong to me at the time; now re-reading the issue based on the examples provided it looks more like the symbol of the `ToolDescriptor` and `GlobalConfiguration` collided, as they are both descriptors in `ToolConfigurationCategory` (and hence grouped in the same top-level `tool` entry in JCasC).

I guess I'm three years late with this, but I retract my objection to the symbol change with this new information 😅

timja:
> My concern is someone else using the same very generic symbols (BOM or not). Based on my assumptions until today, those would collide as they're all in the same namespace.

Symbols should all be scoped to the extension point and JCasC discovers them relative to the configurator. It'll only discover a subclass of `PropertyEntry` in this scope not that jcasc is going to be looking at these properties anyway if I understand correctly

daniel-beck:
> in this scope not that jcasc is going to be looking at these properties anyway if I understand correctly

Right, I was looking at this from the POV of another plugin having a `user`/`group` symbol for something configured by JCasC and that being confused whether it is this, or that plugin's. Thanks for confirming this should work!

sabretus:
My DSL script is broken after these changes, submitted a bug https://issues.jenkins.io/browse/JENKINS-72025

daniel-beck:
Already responded before you commented here.

PR #145, opened 2023-07-04
Nicer JCasC syntax for defining permissions

daniel-beck:
Alternative to #111 with (IMO) nicer structure.

Complements #144 (which is for Job DSL / Pipeline).

```
- group:
    name: foo
    permissions:
    - Overall/Administer
    - …
```

Feedback welcome.


src/test/resources/org/jenkinsci/plugins/matrixauth/integrations/casc/configuration-as-code-v3.yml
```diff
@@ -0,0 +1,95 @@
...
+              'GROUP:Job/Delete:authenticated',
+              'GROUP:Job/Read:authenticated',
+          ])
+          entries([
+            group('authenticated') {
+              'Job/Build',
+              'Job/Configure',
+              'Job/Delete',
+              'Job/Read'
+            }
```
daniel-beck, line 91: Dependency on #144

daniel-beck:
@timja Just in case, any thoughts about this?

timja:
It looks good to me, just some tests to sort out by the looks of it? (a bunch of commented unit tests)

src/test/resources/org/jenkinsci/plugins/matrixauth/integrations/casc/ExportTest/ExportTest-exportTest-node.yml
```diff
@@ -0,0 +1,12 @@
...
+- group: "authenticated"
+  permissions:
+  - "Agent/Build"
+  - "Agent/Configure"
+  - "Agent/Connect"
+  - "Agent/Delete"
+  - "Agent/Disconnect"
+- permissions:
+  - "Agent/Build"
+  user: "anonymous"
```
daniel-beck, line 11: Sorting of keys here makes this look awkward 👎

src/test/java/org/jenkinsci/plugins/matrixauth/integrations/casc/ExportTest.java
```diff
@@ -39,38 +36,11 @@ public void exportTestLegacy() throws Exception {
...
-                        "View/Delete:authenticated");
-                for (CNode entry : permissions) {
-                    String value = entry.asScalar().getValue();
-                    assertTrue("list contains entry " + value, strings.contains(value));
-                }
-            }
+            assertEquals(
+                    toStringFromYamlFile(
+                            this,
+                            "/org/jenkinsci/plugins/matrixauth/integrations/casc/ExportTest/ExportTest-exportTestLegacy-global.yml"),
```
timja, line 42: the classpath entry should be relative, I would have thought `ExportTest-exportTestLegacy-global.yml` would work but this stuff can be a real pain sometimes and very hard to debug

timja (approved):

src/test/java/org/jenkinsci/plugins/matrixauth/integrations/casc/ExportTest.java
daniel-beck: I wanted these files in the folder corresponding to the class, not in the folder for the package. (Kinda redundant with the file name now though 🤔 )

timja: `ExportTest/exportTestLegacy-global.yml` would work then I think?

daniel-beck: Huh. Never tried that. Javadoc indicates this could work.

daniel-beck:
FWIW I've attempted to maintain backward compatibility with older versions of the JCasC syntax (the YAML part, not the Job DSL Groovy part) and it _should_ work with those older configs, but I still recommend caution when updating. The Job DSL and Pipeline syntaxes aren't as configurable and those are fully incompatible.

> Do you know if the breaking change also applies to JCasC settings for the [Azure AD](https://plugins.jenkins.io/azure-ad/) plugin, which depends on matrix-auth? … the azure-ad-plugin incompatibility could deserve a mention in the release notes

With regards to `azure-ad`, [it ignores restrictions declaring certain APIs in this plugin to be inaccessible to other plugins](https://github.com/search?q=repo%3Ajenkinsci%2Fazure-ad-plugin+SuppressRestrictedWarnings&type=code), which allows me to, e.g., freely rename methods, so any problems exclusive to `azure-ad` I do not care about at all. It that plugin's maintainers' responsibility to maintain compatibility. FWIW Tim is aware of these PRs, he provided the inspiration and a first attempt in #111 and some feedback/assistance for what ultimately got integrated. Other than possibly release timing, these changes shouldn't catch him by surprise.

daniel-beck:
@PiersRBME They don't show up on https://plugins.jenkins.io/matrix-auth/dependencies/ as dependants, so do not use _any_ APIs of this plugin, except of those defined in core.

Looks to me like y'all need staging environments 😉

timja:
(in progress for AzureAD)

PiersRBME:
Yeah, have non-prod envs and was able to check and update the CASC format - it is clearer.

But, yeah backwards compat is hard. Someone somewhere will come to rely on everything, private or not. To an outsider a change in one plugin that breaks another is an irritation (at best) -- whether the dev doesn't care about the other at all is not much comfort!

timja:
Not done yet but I've opened a draft PR: https://github.com/jenkinsci/azure-ad-plugin/pull/460

src/main/java/org/jenkinsci/plugins/matrixauth/integrations/casc/MatrixAuthorizationStrategyConfigurator.java
```diff
@@ -30,20 +37,80 @@ public Class<?> getImplementedAPI() {
...
+                        container.add(p, e.permissionEntry());
+                    });
+        });
+    }
+
     /**
      * Extract container's permissions as a List of "TYPE:PERMISSION:sid"
      */
-    public static Collection<String> getPermissions(AuthorizationContainer container) {
+    public static Collection<String> getLegacyPermissions(AuthorizationContainer container) {
```
timja, line 113: method appears to be unused

daniel-beck: Thanks, filed https://github.com/jenkinsci/matrix-auth-plugin/pull/153 and will clean that up in the next release.

PR #152, opened 2023-08-22
Clean up some JS

daniel-beck:

PR #140, opened 2023-05-12
[JENKINS-70906] Compatibility with Prototype removal

daniel-beck:
Based on https://github.com/jenkinsci/matrix-auth-plugin/pull/139, actual changes are from 4th commit onward.

Compatibility with [JENKINS-70906](https://issues.jenkins.io/browse/JENKINS-70906).

Tested with and without https://github.com/jenkinsci/jenkins/pull/7986, including ambiguity-related code and SystemRead / ExtendedRead.

PR #156, opened 2023-10-12
prototype js removal code changes

rsahoo7495:
changed YAHOO.lang.escapeHTML() method to javascript method in js file https://github.com/jenkinsci/matrix-auth-plugin/blob/master/src/main/resources/hudson/security/table.js
### Testing done
mvn verify and mvn clean install got success
After changes and build, added plugin hpi to jenkins instance

image

provided global access to the user for read

image

provided project level access for build and read

image

Now we can have access to in the project for build

image

Debugging in browser with code, it still works well with new code

image

image

image

```[tasklist]
### Submitter checklist
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

### DIFF
--- a/src/main/resources/hudson/security/table.js
+++ b/src/main/resources/hudson/security/table.js
@@ -61,11 +61,9 @@ Behaviour.specify(".matrix-auth-add-button", 'GlobalMatrixAuthorizationStrategy'
   }
 });
 
-  function escapeHTML(text) {
-        var div = document.createElement('div');
-        div.appendChild(document.createTextNode(text));
-        return div.innerHTML;
-    }
+function escapeHTML(html) {
+  return html.replace(/'/g, "&apos;").replace(/"/g, "&quot;").replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+}
 /*
  * Behavior for the element removing a permission assignment row for a user/group
  */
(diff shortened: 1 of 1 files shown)


daniel-beck:
Interesting approach to sanitize the string with the text node. I did it with basic escaping in [#152](https://github.com/jenkinsci/matrix-auth-plugin/pull/152/commits/16fb197cf5b9b879ede52b82e9e15355baa0a12b#diff-c6b2a5410964a2ef01459a94e3469b313dcfb67b62dd32a222ac5715829ab9efR3-R5), but this seems more elegant.

The testing done is unconvincing, it does not actually invoke the code in a manner demonstrating that escaping is still being done correctly.

rsahoo7495:
@daniel-beck thanks for your comment, escaping which u did in in [#152](https://github.com/jenkinsci/matrix-auth-plugin/pull/152/commits/16fb197cf5b9b879ede52b82e9e15355baa0a12b#diff-c6b2a5410964a2ef01459a94e3469b313dcfb67b62dd32a222ac5715829ab9efR3-R5) looks more convenient to me

rsahoo7495:
did testing with input <>"'&Raj and added debugging screenshot

julieheard (approved):

daniel-beck:
In that case I'll close this PR as it doesn't offer anything over the cleanup in #152. Note that I removed Prototype in #140, this just removes uses of YUI. The plugin is compatible with Jenkins 2.426+ already.

PR #158, opened 2024-02-21
Simplified `NodeContributor.record` on updated

jglick:
Simplifies logic introduced in #110. @Vlatombe noticed a test failure from https://github.com/jenkinsci/jenkins/pull/8979 since the logic was assuming that a save of an agent would use `Nodes` as the subject, which is now more naturally the `Node`, but anyway there was already a listener method for this.


Vlatombe (approved):

jglick:
@daniel-beck ping

daniel-beck (approved):

PR #147, opened 2023-07-13
UI renovation

mawinter69:
- use symbols everywhere
- use new button style
- align things and add gaps

Before:
image

1. No gap to border
2. checkboxes not centered
3. icons not centered
4. old style user icon
5. no gap
6. old style group icon
7. no border between permissions
8. no gap to warning message

After:
image

### Testing done

```[tasklist]
### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue
```

PR #161, opened 2024-07-12
Remove outdated comments

zbynek:
This plugin no longer uses the YUI buttons, so mentioning them in a comment is misleading.
(The goal is just to clean up the search results for `makeButton` in the `jenkinsci` org)


daniel-beck:
Thanks!

Amends https://github.com/jenkinsci/matrix-auth-plugin/pull/147 which changed the tag.

PR #165, opened 2024-10-22
Format file with Prettier

basil:
Run `npx prettier --print-width 999 -w jenkinsci/matrix-auth-plugin/src/main/resources/hudson/security/table.js` on this file to make it easier to read.

If we want to drop the `--print-width` option (which results in even more invasive changes) I can do that as well.

If this PR isn't desired, that is fine as well. We can simply close it and leave the status quo; I don't feel strongly about this.

### Testing done

None, but I trust Prettier to not introduce new bugs.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,2 +1,6 @@
 # mvn spotless:apply
 4d0e23d0861ee54bcf40dc3070969584c4eb2de5
+
+# https://github.com/jenkinsci/matrix-auth-plugin/pull/165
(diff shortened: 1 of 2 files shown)


daniel-beck:
Thanks and sorry for the delayed response. I kept wondering whether this is going to be useful long term, and for me it comes down to whether this is in scope for being enforced by Spotless or similar during the build. I guess not since you're only changing this file?

basil:
Yes, I am only changing this file.

PR #163, opened 2024-08-30
Build against core PR with ee9

daniel-beck:
### Submitter checklist
- [ ] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [ ] Ensure that the pull request title represents the desired changelog entry
- [ ] Please describe what you did
- [ ] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #167, opened 2025-01-16
Migrate from EE 8 to EE 9

basil:
### Testing done

`mvn clean verify` both here and in `azure-ad-plugin` (with these changes)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>4.88</version>
+    <version>5.5</version>
   </parent>
   <artifactId>matrix-auth</artifactId>
   <version>${revision}${changelist}</version>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -28,16 +28,18 @@
...
   </properties>
 
   <dependencyManagement>
     <dependencies>
       <dependency>
         <groupId>io.jenkins.tools.bom</groupId>
-        <artifactId>bom-2.426.x</artifactId>
-        <version>3208.vb_21177d4b_cd9</version>
+        <artifactId>bom-${jenkins.baseline}.x</artifactId>
+        <version>3893.v213a_42768d35</version>
```
daniel-beck, line 42: Has https://github.com/jenkinsci/cloudbees-folder-plugin/pull/444 which addresses https://github.com/jenkinsci/matrix-auth-plugin/pull/163#issuecomment-2321505932

src/main/java/org/jenkinsci/plugins/matrixauth/AuthorizationPropertyDescriptor.java
```diff
@@ -48,8 +50,34 @@ public interface AuthorizationPropertyDescriptor<T extends AuthorizationProperty
 
     T create();
 
+    default T createNewInstance(StaplerRequest2 req, JSONObject formData, boolean hasOptionalWrap)
```
daniel-beck, line 53: Why? This class is `@Restricted(NoExternalUse.class)`.

daniel-beck: AFAICT at this point `azure-ad` only reuses inheritance types from `matrix-auth`, IIRC since the last time I broke something there by changing `@Restricted` code here.

@timja Could you confirm?

daniel-beck (approved):

PR #171, opened 2025-06-14
Support "Microsoft Entra ID"-plugin by displaying readable group names.

cvberg:
When using this plugin in combination with the "Microsoft Entra ID" plugin, users specified using an OID
will be shown with a readable name. However, groups do not get the same treatment and will always show an OID.  
This PR will solve that issue.  
It's a literal copy of a change made to the "role-strategy-plugin": https://github.com/jenkinsci/role-strategy-plugin/pull/355/files  
Jenkins issue: https://issues.jenkins.io/browse/JENKINS-75726  

Using OIDs is unfortunately the only way to make the Microsoft Entra ID plugin and this plugin work together. This PR will make this limitation less problematic.

### Testing done

- I've tested this with the Entra ID plugin installed and the "Azure AD" authentication enabled. Before groups entered with a OID, will now show their Display Name instead.
- I've also tested it with Active Directory authentication enabled. For AD, OIDs are not supported, so you must specify the group using the Display Name and that name will be shown unchanged.
- I've repeated the previous test with the Entra ID plugin removed.

These tests have been done manually because I don't know how to add automated tests without having access to a real Entra ID system. I tested this at my workplace, where we do have this infrastructure.

### Submitter checklist
- [X] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [X] Ensure that the pull request title represents the desired changelog entry
- [X] Please describe what you did
- [X] Link to relevant issues in GitHub or Jira
- [X] Link to relevant pull requests, esp. upstream and downstream changes
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/matrixauth/ValidationUtil.java
+++ b/src/main/java/org/jenkinsci/plugins/matrixauth/ValidationUtil.java
@@ -31,6 +31,7 @@ import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Functions;
 import hudson.Util;
 import hudson.model.User;
+import hudson.security.GroupDetails; // PATCH
 import hudson.security.SecurityRealm;
 import hudson.security.UserMayOrMayNotExistException2;
 import hudson.util.FormValidation;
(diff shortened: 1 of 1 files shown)


cvberg:
Sorry, I made a mess of authorship in git. I'll prepare a new PR.
BTW: the people creating the Entra ID plugin.made the wrong choice when implementing permissions. They chose "matrix" instead of "project matrix". The latter supports inheritance, while the former doesn't.

PR #173, opened 2025-07-08
Wait a few seconds for test cleanup on Windows testWidgets

MarkEWaite:
## Wait a few seconds for test cleanup on Windows testWidgets

The testWidgets() test fails consistently on ci.jenkins.io since the transition from Azure to AWS.  Since the tests are already using long sleep calls for various other purposes, let's sleep at the exit from the failing test so that there is time for Java to close files.

### Testing done

Confirmed that `mvn clean verify` passes on my Linux computer.  Confirmed on ci.jenkins.io that [Windows tests pass](https://ci.jenkins.io/job/Plugins/job/matrix-auth-plugin/job/PR-173/lastSuccessfulBuild/testReport/).

### Submitter checklist
- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your main branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

PR #174, opened 2025-07-08
Wait before exiting testWidgets

MarkEWaite:
## Wait before exiting testWidgets

The `testWidgets` test has failed 2 out of 2 builds on the master branch since the transition of ci.jenkins.io from Azure to AWS.   Other plugins have seen similar behavior and have reduced the failures significantly by waiting several seconds at the end of the test before exiting the test.

Pull requests that have used that technique include:

* https://github.com/jenkinsci/config-file-provider-plugin/pull/381
* https://github.com/jenkinsci/dashboard-view-plugin/pull/417

A 3 second sleep was enough to pass 3 of 4 test runs in pull request:

* https://github.com/jenkinsci/matrix-auth-plugin/pull/173

### Testing done

Confirmed that the tests pass locally.  Confirmed that the tests pass on ci.jenkins.io 9 of 9 times.  Ready for review

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/matrixauth/Security2180Test.java
+++ b/src/test/java/org/jenkinsci/plugins/matrixauth/Security2180Test.java
@@ -294,8 +294,8 @@ public class Security2180Test {
         final String contentAsString = htmlPage.getWebResponse().getContentAsString();
         assertThat(contentAsString, not(containsString("job/folder/job/job"))); // Fails while unfixed
         if (Functions.isWindows()) {
-            // Wait a few seconds so that files are closed before test cleanup
-            Thread.sleep(3031);
+            // Wait several seconds so that files are closed before test cleanup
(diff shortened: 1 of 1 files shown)


daniel-beck:
Seems like something that should be done in JTH, no? Or are the conditions so unusual that this is not widespread?

MarkEWaite:
> Seems like something that should be done in JTH, no? Or are the conditions so unusual that this is not widespread?

It is not widespread as far as I can tell.  From the 100+ plugins that I checked as part of Monday's transition from Azure to AWS, I found less than five cases that needed this type of change.  Two of those cases were specifically related to the Maven project type.  I don't know the details of the conditions that cause the failure and was unwilling to spend the time to investigate those details because there were so few failing cases among so many tests.

PR #190, opened 2026-04-02
Redesign UI

timja:
TODO:
* [x] Filter box would be nice if the border radius's were appropriate, right now they don't have when others around them get hidden
* [x] ATH - passing now, PR needs merging after this is released
  * https://github.com/jenkinsci/acceptance-test-harness/pull/2665
* [x] PCT - passed
  * https://github.com/jenkinsci/bom/pull/6563

Part of https://github.com/jenkinsci/sig-ux/issues/10

_Future enhancement: Once wizard functionality is in core we could look to improve the add user and group flow so that users don't get returned to the page and have to select permissions for a new entry they could do that in a flow which I think would be nicer._

Redesigns the plugin interface, see video:

https://github.com/user-attachments/assets/c15b8fc1-501f-45b7-9411-012e811a5507

Screenshots:

<details>
<summary>Simple</summary>

image

</details>

<details>
<summary>Expanded</summary>

image

</details>

<details>
<summary>Ambiguous</summary>

image

</details>

This was originally written by Gen AI with direction from myself. Jan and I tidied up the code and reviewed every line

### Testing done

* Tested project and global auth strategies
  * Test at project, folder, system level
* Ambiguous entries
  *  migrating them as well

### Submitter checklist
- [ ] Ensure you have provided tests that demonstrate the feature works or the issue is fixed


timja:
cc @janfaracik, 

fyi @mawinter69, this is what I meant about doing a redesign of the security plugins so the tables aren't a problem with the sidepanel.

mawinter69:
In role-strategy that will work for the page where you define the roles. For the assign page, that might require some adjustments. When you have 100 or even 300 roles and want to assign roles to users that will make the opening area below really huge and lead to scrolling. 
Currently role-strategy has 2 filters on the assign page so you can make the matrix small and manageable. So something like that is also needed in a new UI.
Another think to consider is that you can't send a checkbox for everything. When you have 300 roles and 300 users you get a form with almost 100k fields. That is over all default limits (10k number of form fields and 200k content size). So what role-strategy does is creating a json that contains only the checked fields as usually the matrix is quite sparse.

mawinter69:
How does that look when you have selected like 10 different permissions, e.g. when you have more plugins that contribute additional permissions

timja:
> How does that look when you have selected like 10 different permissions, e.g. when you have more plugins that contribute additional permissions

The summary only shows what there is room for an then there is ellipses to show there's more:

image

image

daniel-beck:
From a casual look, this seems great, but would be interesting to get an idea of the scale of some instances' configuration first, as this doesn't look particularly scalable to dozens of users/groups. My hunch is that there are some crazy configurations out there.

It also looks impossible to find who has a given permission, at least if I'm reading the JS right. Right now I highlight a table column and can easily see it, here it looks like I'd need to expand all sections and deal with a much more complex grid.

Re "grant everything" only giving Administer instead of everything, does that work with optional permissions not implied by anything else, like Run/Artifacts?

mawinter69:
> Re "grant everything" only giving Administer instead of everything, does that work with optional permissions not implied by anything else, like Run/Artifacts?

Seems not. Any reason why it was designed like that? Is that maybe a leftover from things like `RunScripts`? So maybe when you pass null as impliedBy then falling back to admin might make sense?

daniel-beck:
> Any reason why it was designed like that?

I would need to dig.

> Is that maybe a leftover from things like `RunScripts`?

Those came later.

> maybe when you pass null as impliedBy then falling back to admin might make sense

Maybe. There may be weird edge cases here as well. Maybe it's just time to clean up those permission declarations; meanwhile the button should do what it says IMO.

timja:
> Re "grant everything" only giving Administer instead of everything, does that work with optional permissions not implied by anything else, like Run/Artifacts?

It works fine from what I can tell - where are you seeing grant admin only? Its just implied don't get checked


https://github.com/user-attachments/assets/30ce58bb-e23b-4534-a77e-2966f9101d7b

daniel-beck:
@timja Great! The existing UI did look different between implied not granted and implied granted, yours didn't seem to.

timja:
> It also looks impossible to find who has a given permission,

Should be possible now since https://github.com/jenkinsci/matrix-auth-plugin/pull/190/commits/220a8329c9072234d9d6f2bb61bb89083a81f454

I tested it quite a lot and I think its setup well. I've made it so that if you search for a permission that they have as implied they'll come back, e.g search read and administer will show.


https://github.com/user-attachments/assets/cd53e562-7de4-4806-a8bd-70df42d5e396



> as this doesn't look particularly scalable to dozens of users/groups

By scalable do you think as it takes up a bit more room? Or something else?

daniel-beck:
> By scalable do you think as it takes up a bit more room? Or something else?

Mostly when it comes to getting an "overview" of permissions -- right now, it's a not a great UI, but it's using space efficiently. Expanding all list entries provides the same information in the new UI, and that's just impractical. Filtering for permissions seems like a good addition though (untested), perhaps that's enough.

timja:
> Future enhancement: Once wizard functionality is in core we could look to improve the add user and group flow so that users don't get returned to the page and have to select permissions for a new entry they could do that in a flow which I think would be nicer.

Thinking: could probably just adding it to the dialog that comes up

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -55,160 +54,236 @@ THE SOFTWARE.
...
-    <local:isEditable>
-      <div class="jenkins-!-margin-top-2 jenkins-buttons-row jenkins-buttons-row--equal-width">
-        <button type="button" class="jenkins-button matrix-auth-add-button" id="${id}UserButton"
-              data-template-id="${id}"
-              data-table-id="${strategyid}"
-              data-type="USER"
-              data-type-label="${%user}"
-              data-message-prompt="${%userPrompt}"
-              data-message-title="${%Add user}"
-              data-message-error="${%userError}">${%Add user…}</button>
```
daniel-beck, line 194: Why remove the `…` suffix that indicates no immediate action is taken on button click?

timja: They don’t mean that to me and i can’t see that here https://en.wikipedia.org/wiki/Ellipsis

they aren’t used elsewhere in Jenkins AFAIK. I think they may have been in advanced awhile ago but they were removed

daniel-beck: > They don’t mean that to me and i can’t see that here https://en.wikipedia.org/wiki/Ellipsis

https://en.wikipedia.org/wiki/Ellipsis#More_input_will_be_needed

Check pretty much any application's menu items, and you should see this pattern.

timja: Ok I see apple uses it fairly consistently, are we using it elsewhere in Jenkins?

cc @janfaracik

janfaracik: I've considered adding it before, but it's not a pattern we're using (at least consistently) in Jenkins, which would just cause more inconsistency.

daniel-beck: > apple

Basically all OSes:

https://learn.microsoft.com/en-us/windows/win32/uxguide/text-ui

> **Ellipses mean incompleteness.** Use ellipses in UI text as follows:
> * …
> * **Commands:** Indicate that a command needs additional information. Don't use an ellipsis whenever an action displays another window only when additional information is required. For more information, see [Command Buttons](https://learn.microsoft.com/en-us/windows/win32/uxguide/ctrl-command-buttons).

also https://learn.microsoft.com/en-us/windows/win32/uxguide/ctrl-command-buttons

> While command buttons are used for immediate actions, more information might be needed to perform the action. **Indicate a command that needs additional information (including confirmation) by adding an ellipsis at the end of the button label.**

https://develop.kde.org/hig/text_and_labels/#ellipses

> End an action’s label with an ellipsis if it always requires additional user input before it completes. This is common for actions that open a dialog prompting the user to make a further choice.
> 
> Bad: Save As
> 
> Good: Save As…

https://developer.gnome.org/hig/guidelines/writing-style.html

> Use an ellipsis (…) at the end of a label if further input or confirmation is required from the user before the action can be carried out. For example:
> 
> * Save As…
> * Find…
> * Delete…

It's less consistent on the web, of course.

timja: The point is about Jenkins if its not consistently used in Jenkins though is it worth having it in just this one place?

daniel-beck: Fair, although ideally we could get there in the future. Just pointing out it's very common, and not just an Apple thing.

daniel-beck:
Overall this looks pretty solid, thanks! I identified a few minor issues so far, and found a few problems:

* Problem: As each (collapsed) row takes more space, and the expanded row takes much more space, it's weirder now than it was in the past that new rows appear at the botton of the list. It is fairly likely to need to scroll to configure a new entry (distance "Add user" button and "Metrics" permissions category in the new entry). Would it be better to add new entries on top? (Related, would then make more sense to sort authenticated and anonymous at the bottom?)
* Bug: Migrating an ambiguous entry does not clear the tooltip and text color of being ambiguous.
* Bug: Migrating an ambiguous "authenticated" entry (and probably "anonymous" too) while filters are live will end up hiding the migrated entry if migrating to a currently filtered out entry.

Other thoughts:

* Why is the maximum width fairly low? Wouldn't it make more sense to use the full width, as the rest of the form does? This may even allow putting permission group titles on the same line as individual permissions.
* The filter field is kinda weird because the text part is names, the button part is permissions
* Should filtering by permissions have "OR" (current behavior) or "AND" semantics?
* Would it make sense to make the "dropdown with checkboxes" a default UI element? Seems a little much custom CSS + JS for me to maintain 😬

The wording should be updated: The description for ambiguous permissions refers to a "table" when there's no table anymore. (Also, did I ever re-read this after writing? The text is terrible. Happy to take this one myself. The plugin uses long obsolete "User ID" terminology cf. recent PR by @janfaracik in core, so I'd need to follow up anyway.)

src/main/resources/hudson/security/GlobalMatrixAuthorizationStrategy/config.jelly
```diff
@@ -55,160 +54,236 @@ THE SOFTWARE.
...
+                               data-implied-by-list="${descriptor.impliedByList(p)}"
+                               data-permission-id="${p.id}">
+                          <j:choose>
+                            <j:when test="${isChecked}">
+                              <input type="checkbox" name="[${p.id}]" checked="checked"/>
+                            </j:when>
+                            <j:otherwise>
+                              <input type="checkbox" name="[${p.id}]"/>
+                            </j:otherwise>
+                          </j:choose>
```
mawinter69:
```suggestion
                          <input type="checkbox" name="[${p.id}]" checked="${isChecked?'checked':null}"/>
```

mawinter69 (changes requested):

timja:
> Overall this looks pretty solid, thanks! I identified a few minor issues so far, and found a few problems:
> 
> * Problem: As each (collapsed) row takes more space, and the expanded row takes much more space, it's weirder now than it was in the past that new rows appear at the botton of the list. It is fairly likely to need to scroll to configure a new entry (distance "Add user" button and "Metrics" permissions category in the new entry). Would it be better to add new entries on top? (Related, would then make more sense to sort authenticated and anonymous at the bottom?)

Sure

> * Bug: Migrating an ambiguous entry does not clear the tooltip and text color of being ambiguous.

Fixed

> * Bug: Migrating an ambiguous "authenticated" entry (and probably "anonymous" too) while filters are live will end up hiding the migrated entry if migrating to a currently filtered out entry.

I think fixed, didn't fully understand this though


> Other thoughts:
> 
> * Why is the maximum width fairly low? Wouldn't it make more sense to use the full width, as the rest of the form does? This may even allow putting permission group titles on the same line as individual permissions.

No reason, I've made it use full width, I think better on its own line, the Job group has too many permissions

> * The filter field is kinda weird because the text part is names, the button part is permissions

I've updated the tooltip to help a bit here

> * Should filtering by permissions have "OR" (current behavior) or "AND" semantics?

OR makes sense to me if I want to build a list of certain permissions

> * Would it make sense to make the "dropdown with checkboxes" a default UI element? Seems a little much custom CSS + JS for me to maintain 😬

I think this is quite niche although useful in this case.

> The wording should be updated: The description for ambiguous permissions refers to a "table" when there's no table anymore. (Also, did I ever re-read this after writing? The text is terrible. Happy to take this one myself. The plugin uses long obsolete "User ID" terminology cf. recent PR by @janfaracik in core, so I'd need to follow up anyway.)

Updated - I did the couple of other places that use that terminology as there was so few.

timja:
reminder @daniel-beck

daniel-beck:
Thanks for the reminder, and again apologies for the delay here.

I tested the current state of the PR and it looks really solid. Still not sure about the piles of extra CSS, but I figured out how to connect Claude to Playwright so that should be fine :trollface: 

In the interest of a quick followup response if things go unexpectedly sideways, I plan to merge and release on Monday.

daniel-beck (approved):

daniel-beck:
I'm running into problems with Spotbugs failing release builds, so that's a little delayed. #191 did not help.

timja:
switch to CD? Can set the prefix to 3.x so the version doesn't jump super high.

daniel-beck:
Not interested in doing that as a workaround for a broken toolchain. Looks like some pointless updates to m-r-p need to be turned back.

pom.xml
```diff
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>5.22</version>
+    <version>6.2138.v03274d462c13</version>
```
daniel-beck, line 8: @timja What motivated this change in the context of this PR?

timja: IIRC to be able to test against recent core versions as something doesn’t support Java 21 in 5.x lib-version-number potentially

